Issue #2, opened 2016-05-03
[cbor] Add support for upcoming packed arrays tags


cowtowncoder:
(originally by neothemachine)

-----

There is a draft RFC to be published soon which adds support for packed/typed arrays in CBOR, see the [current version](https://tools.ietf.org/html/draft-jroatch-cbor-tags-02). As one of the original CBOR authors is also an author of this proposal I'm sure it will gain widespread adoption in encoders and decoders, especially because it is such a useful feature and will help greatly when big packed arrays need to be efficiently encoded and also decoded both in terms of space usage and speed.

I thought it would be good putting it on the radar for implementation. I think it will be rather straightforward to implement. I currently don't have time for it myself in the next 3-4 weeks but I certainly will help where I can, reviewing code etc.

cowtowncoder:
Quick note: Should be (more) doable with Jackson 2.8, planning to add `JsonParser` support to make it possibly to efficiently and simply support a small set of "well-known" array types; `int[]` and `long[]` at least, but maybe `short[]` for good measure.

Could be https://github.com/FasterXML/jackson-core/issues/241 or perhaps something even more direct.

cowtowncoder:
So, explicit array writes are now included for `int[]`, `long[]` and `double[]`: if packed arrays functionality becomes official CBOR feature it should be much easier to support those now.

Issue #1, opened 2016-05-03
[cbor] Implement `JsonGenerator.writeStartArray(int)`


cowtowncoder:
(moved from https://github.com/FasterXML/jackson-dataformat-cbor/issues/1)

When `jackson-core` adds the new variant of `writeStartArray`, CBOR module should implement it, at least for cases where length is small enough and we can save one byte from encoded length. Ideally we should also verify that actual number of value writes matches to indicated length.

cowtowncoder:
Duplicate of #16, fixed via #23; included in 2.8.0. Closing as duplicate.

PR #23, opened 2016-06-27
Support sized array and sized map.

mbaril:
*Added support for write field name with long type values.
 This is to be able to create map with number as key.

*Added support for size arrays and maps

*Fix CBOR Parser bug when fieldname is not a string.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
@@ -135,13 +135,19 @@ public class CBORGenerator extends GeneratorBase {
     protected boolean _cfgMinimalInts;
 
     /**
-     * List that contains the remaining elements of all arrays and objects.
-     * When unsized arrays or maps are added in the list,
-     * the number of elements is set to a specific value : INDEFINITE_LENGTH_TAG
+     * Special value that is use to keep tracks of arrays and maps opened with infinite length
      */
-    protected final List<Integer> listOfSizes = new ArrayList<Integer>();
+    private final int INDEFINITE_LENGTH = Integer.MIN_VALUE -1;
 
-    private final int INDEFINITE_LENGTH_TAG = Integer.MIN_VALUE;
+	/**
+	 * List that contains the number of remaining elements for parents arrays and objects.
+	 */
+	protected final List<Integer> remainingElementsList = new ArrayList<Integer>();
+	
+	/**
+	 * Number of elements remaining in the current complex structure (if any). 
+	 */
+	private int currentRemainingElementsCount = INDEFINITE_LENGTH;
 
     /*
      * /********************************************************** /* Output
@@ -382,7 +388,7 @@ public class CBORGenerator extends GeneratorBase {
         if (_writeContext.writeFieldName(name) == JsonWriteContext.STATUS_EXPECT_VALUE) {
             _reportError("Can not write a field name, expecting a value");
         }
-        _decrementList();
+        decrementElementsRemainingCount();
         _writeString(name);
     }
 
@@ -399,7 +405,7 @@ public class CBORGenerator extends GeneratorBase {
             _writeByte(BYTE_EMPTY_STRING);
             return;
         }
-        _decrementList();
+        decrementElementsRemainingCount();
         _writeLengthMarker(PREFIX_TYPE_TEXT, len);
         _writeBytes(raw, 0, len);
     }
@@ -417,7 +423,7 @@ public class CBORGenerator extends GeneratorBase {
             return;
         }
         _verifyValueWrite("write String value");
-        _decrementList();
+        decrementElementsRemainingCount();
         _writeString(value);
     }
 
@@ -426,7 +432,7 @@ public class CBORGenerator extends GeneratorBase {
         if (_writeContext.writeFieldName(String.valueOf(size)) == JsonWriteContext.STATUS_EXPECT_VALUE) {
             _reportError("Can not write a field name, expecting a value");
         }
-        _decrementList();
+        decrementElementsRemainingCount();
         _writeNumberNoCheck(size);
     }
 
@@ -479,8 +485,8 @@ public class CBORGenerator extends GeneratorBase {
         _writeContext = _writeContext.createChildArrayContext();
         _writeByte(BYTE_ARRAY_INDEFINITE);
 
-        _decrementList();
-        listOfSizes.add(INDEFINITE_LENGTH_TAG);     // set an unsized tag in the list for this new array
+        decrementElementsRemainingCount();
+        openComplexElement(INDEFINITE_LENGTH); // set an unsized tag in the list for this new array
     }
 
     /*
@@ -495,8 +501,8 @@ public class CBORGenerator extends GeneratorBase {
 
         _writeLengthMarker(PREFIX_TYPE_ARRAY, size);
 
-        _decrementList();
-        listOfSizes.add(size);
+        decrementElementsRemainingCount();
+        openComplexElement(size);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -817,7 +817,7 @@ protected String _numberToName(int ch, boolean neg) throws IOException
         if (neg) {
             i = -i - 1;
         }
-        return String.valueOf(1);
+        return String.valueOf(i);
```
cowtowncoder: I will go ahead and fix this separately, backporting in 2.7 at least.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -402,9 +417,20 @@ public final void writeStringField(String fieldName, String value)
             return;
         }
         _verifyValueWrite("write String value");
+        _decrementList();
         _writeString(value);
     }
 
+	
+    public final void writeFieldLong(long size) throws IOException {
```
cowtowncoder: So, would the intent here be to allow writing long-keyed Maps/Objects? Since this would not be exposed via Jackson streaming API, the only way to use this would be with custom serializers. But at very least Javadocs would have to explain the method (including where it was added); this would have to be in 2.8 as it's new part of API.
I am not 100% sure if it makes sense to add format-specific extensions, but I guess that could be acceptable, as some format modules have something similar (I think CSV at least, perhaps XML).

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder: Fix checked in 2.7 (and actually even 2.6 although not sure there will be new releases from that branch)

cowtowncoder:
Ok. First things first: thank you for contributing this patch! I know that ability to use length-prefixed objects and especially arrays is something that has been requested.

I went ahead and merged simplest part of the patch (fixing type of `1` that should be `i`), since that can be merged into earlier versions.

Of remaining changes, I am bit ambivalent regarding addition of ability to write `long` keys: it may be acceptable to do that, but I'll have to think about it a bit. There's always bit of challenge in dealing with format-specific parts, and I am not sure what the right balance is here.

And then the bulk of changes: supporting length-prefixed writes. At logical level, I think implementation makes sense and is correct. But I have concerns about efficiency of the implementation. Given that checks are called for each and every value write, and that they are using JDK provider `List` with `Integer` wrappers, I would expect this to have measurable detrimental effect on performance compared to existing code. I hope to find time to test this locally, using `jackson-benchmarks`.

But I think implementation could be improved to mostly/completely eliminate the overhead. Two ways to go about it that I can think of are:

1. Use simple `int[]` instead of `List`; code may be bit uglier but access much more efficient. Could alternatively create helper object to contain functionality, essentially similar to primitive-valued stack that a lib like hppc or alternatives would give (but that we can't use due to sheer size of these libs -- I'd love to use them but don't think it's appropriate).
2. Use custom `JsonWriteContext` to contain per-level counts; this would just be one new field, no List/Stack needed.

and actually either way it would probably make sense to have one new field in generator itself for "current count", so that regular value writes would only modify that count and not look up state from `List`.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
mbaril: We are implementing a protocol that use number as key in Maps. 
For more information: http://core-wg.github.io/yang-cbor/draft-veillette-core-cool-latest.html 

I understand that this method won't be exposed via the streaming API but in our case, we are currently using the CBORGenerator directly (we are not using the ObjectMapper). 

Do you think it could be feasible to modify the jackson-databind project to provide such flexibility? Using annotations for example.

mbaril:
For the performance related concern:

For  #1 I agree that there will be a performance impact but it's difficult to tell without testing if the effect will be negligible or not. I was not aware of the project jackson-benchmarks and thus I haven't tried to measure the performance impacts of the modifications.

For #2, this is something that I look at initially but I had some doubt that is was the right way to go when I looked at how that would be implemented. (Extending JsonWriteContext, overriding some functions, ...)

Concerning your last proposition (use a currentCount) I think that this is something that could be easily implemented and that could also help to reduce the performance impact.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
cowtowncoder: I don't think annotations would be needed, as type information itself is available wrt Maps (as POJO fields always have String names). So it'd be adding new methods in `JsonGenerator` (and possibly `JsonParser`), then wiring.
But realistically speaking I don't think this would be implemented by core team any time soon.

cowtowncoder:
I agree that use of `JsonWriteContext` may get tricky. For what it is worth, custom write context is used by `jackson-dataformat-avro` and `jackson-dataformat-protobuf` (both now under `jackson-dataformats-binary` repo), although I think neither tracks counts; protobuf does keep track of content length indirectly.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
mbaril: As I said, from my perspective only, this is not something that is required (as I use the CBORGenerator directly).

I only need to be able to create map and be able to use number as key.

cowtowncoder:
@mbaril Actually, thinking bit more about possibly exposing non-String keys, I think it might be a good idea to start exposing this via `JsonGenerator` interface, perhaps first only adding method for `long` keys (which can generally be used for `int`, `short` keys too), and once in place, can be possibly pipe through from databind too. This could be implemented as simple coercion to `String` in `JsonGenerator` itself, overridden by backends that do support such key types.

The only question really is timing: I am trying hard to finalize 2.8.0. But. I think addition of maybe `writeFieldId(long)` (instead of `writeFieldName()`) should be fine and doable for 2.8.0 -- additional work could then be done for 2.9.0.

I will create an issue at `jackson-core` for this.

mbaril:
Sound perfect. Thanks!

I also took time too run the jackson-benchmarks to compare commit 2.8.0.rc2-2-g08854c6 with 2.8.0.rc2-4-gb6a63b9.

I haven't look the tests in details but it looks like the modifications have an impact on performance.
(I have run the benchmark with the options com.fasterxml.jackson.perf.cbor.* -wi 7 -i 7 -f 3 -t 2)

[benchmarks_results.zip](https://github.com/FasterXML/jackson-dataformats-binary/files/338025/benchmarks_results.zip)

cowtowncoder:
@mbaril Thank you for running the tests and making the change for current count. From simple runs I did locally I could not see much of performance difference, nor was there anything in profiler, so unless I made a mistake it looks like patch as-is wouldn't add measurable overhead for usage where count functionality is not used. I may want to do some minor tweaking (for naming etc), perhaps encapsulate `List` access in separate class, but all in all I think things look good.

It also looks like we already have CLA for you which is good.

So I think what I'll do is merge the patch, and get this included in 2.8.0. Thanks!

cowtowncoder:
Interesting. With some minor playing with attempts to optimize, found out that there is one very performance-sensitive part in `decrementElementsRemainingCount()`.
I was trying to add better checking to ensure that the first "extra" element would trigger exception (and not just when closing the scope which happens well after), but additional check seemed to have disproportionate effect, possibly due to inlining (or lack thereof).

On plus side, patch as-is is remarkably nicely behaving at least wrt `jackson-benchmarks`, so that's good.

PR #27, opened 2016-08-12
Fixed long deserialization problem for longs of ~13digit length

shotbythought:
The decoding for variable length long should use the previously calculated part into account otherwise it throws away some of the long during deserialization. Included is a test that reproduces the error given the previous code.


cowtowncoder:
Makes sense. Thank you for contributing this!

cowtowncoder:
Fixed for 2.8.2, but also backported in 2.7.7 release. Thanks again!

shotbythought:
No problem! Thanks for responding so quickly.

PR #36, opened 2017-01-12
Fix enum handling for Avro 1.8

marcusb:
The GenericDatumWriter in Avro 1.8 became more strict about enums. It
no longer accepts raw enum values, those must be wrapped in a
GenericEnumSymbol (cf AVRO-1810 and AVRO-1870).

Although we still depend on Avro 1.7.7, this change is needed for
projects that use Avro 1.8.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/ser/NonBSGenericDatumWriter.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/ser/NonBSGenericDatumWriter.java
@@ -6,6 +6,7 @@ import java.util.List;
 
 import org.apache.avro.Schema;
 import org.apache.avro.Schema.Type;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericDatumWriter;
 import org.apache.avro.io.Encoder;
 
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Ok thank you for submitting this. I assume it works fine with 1.7.7 still.

One (and only) piece of process before I can merge this (I'll try to apply it against 2.8 branch): a CLA is needed. Individual one:

is most often used; although corporate users sometimes opt for CCLA (https://github.com/FasterXML/jackson/blob/master/contributor-agreement-corporate.txt).
The usual way is to print the doc, fill & sign, scan and email to `info` at fasterxml dot com.

If you could do that, I'll be happy to merge this fix!

Thank you once again.

marcusb:
It works with 1.7.7 because I didn't bump the avro dependency in the POM, but it also works if it's replaced with 1.8.0.

I'll look into the corporate CLA.

cowtowncoder:
Much appreciated wrt 1.7.7 vs 1.8.0 -- I am not sure what would be optimal time to upgrade (and what is level of compatibility), so for immediate future forwards-compatibility is great.

marcusb:
@cowtowncoder Submitted CLA to clas@.

PR #42, opened 2017-02-09
Added maven dependency to main readme

LuboVarga:
Just added latest maven dependency to main readme. It will make it easier to use project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,15 @@ Currently included are:
 * [Protobuf](protobuf/)
 * [Smile](smile/)
 
+## Maven dependency
+```
+<dependency>
+    <groupId>com.fasterxml.jackson.dataformat</groupId>
+    <artifactId>jackson-dataformats-binary</artifactId>
+    <version>2.8.6</version>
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Um, please explain your intent here. Why would anyone ever depend on `jackson-dataformats-binary`?

LuboVarga:
I have expected that adding dependency to parent project will bring me also all modules (protobuf, avro,...). That was wrong assumption. This PR should be closed probably.

Suggestion for particular modules is to add to its read me page dependency to given module. It would make it easier for user to copy/paste and start using it...

cowtowncoder:
@LuboVarga Ok yes. In this case it is simply parent pom for convenience of releasing all at same time.

I'd be happy to accept PR for alternative change(s); not sure if main-level inclusion or module-level make more sense. Problem for me is simply that I don't know of a good way to templatize things to show the latest stable release, so manual updates are often needed. This becomes non-trivial overhead over 20+ modules Jackson has.

PR #44, opened 2017-02-17
[AVRO] Add support for native typeIDs

baharclerode:
This PR implements support for native type IDs for the `jackson-dataformat-avro` module, and fixes numerous bugs related to type IDs that affect interoperability with the Apache reference implementation of Avro.

Also included is the beginnings of a comprehensive test suite to validate that a default `AvroMapper` is compatible with the Apache implementation in schema generation, serialization, and deserialization. Over 350 new unit tests verify that POJOs can be successfully round-tripped in the following combinations:

* Jackson -> Jackson with Jackson schema
* Apache -> Jackson with Jackson schema
* Jackson -> Apache with Jackson schema
* Jackson -> Jackson with Apache schema
* Apache -> Jackson with Apache schema
* Jackson -> Apache with Apache schema

**Note:** While this PR is in a mergable _state_, I still have additional tests I still want to add before it is actually _merged_. I'm opening this to get initial feedback from maintainers and be told all the places I'm not following the style conventions 🙃 . (Most notably, still need to finish implementing annotation support and tests for `@AvroEncode` and `@AvroIgnore`, and support for logical types)

## Bugs Fixed
* [Parser] Fixed bug where reading a float consumed 8 bytes from input instead of 4 and also read incorrect value (`AvroParserImpl.java:233`)
* [Parser] Fixed bug where avro-encoded characters were decoded as ints (`AvroReaderFactory.java:86`)
* [Schema/Generator] Fixed bug where `char[]` schema is array of strings instead of array of int with native type ID of `java.lang.Character` (`ArrayVisitor.java:43`) (Added generator feature to revert to old behavior)
* [Schema] Fixed namespaces for nested classes not including enclosing class name and `$` (`AvroSchemaHelper.java:16`)
* [Schema] Fixed `float` and `long` having an `Type.DOUBLE` schema (`AvroSchemaHelper.java:57`)
* [Schema] Fixed stringable-but-not-string map keys not including key type ID (`MapVisitor.java:53`)
* [Schema] Fixed schema for `byte[]` not including native type ID (`VisitorFormatWrapperImpl.java:111`) 
* [Generator] Fixed unable to serialize Maps with Array or List values (`MapWriteContext.java:39`)
* [Generator] Fixed unable to serialize strings or `char[]` when schema is `char[]` (`NonBSGenericDatumWriter:72`, `ObjectWriteContext.java:99`, `RootContext.java:99`)
* [Generator] Fixed unable to serialize strings or `char[]` when schema is nullable `char[]` (`NonBSGenericDatumWriter.java:43`)

## Features Added
 * [Parser] Support for reading Avro native type IDs (embedded in schema)
 * [Schema] Support for emitting Avro native type IDs (embedded in schema)
 * [Schema] Generator Feature for encoding `char[]` as strings (default: `false`)


cowtowncoder:
First of all, thank you for the contribution! It will take a while for me to read the code; I'll be adding notes as I go.
I especially appreciate improved test suite.

Next one question: would it be possible to base the PR on 2.8 branch instead? I guess I could manually backport at least some, but perhaps it'd be easy for you to rebase (although there has been some refactoring so there may be challenges).
I hope to get 2.9.0.pr1 out within couple of weeks so this may not be a big deal, but `2.8.7` will be released and available earlier.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
```diff
@@ -0,0 +1,238 @@
...
+import org.apache.avro.reflect.CustomEncoding;
+import org.apache.avro.util.Utf8;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+
+public class AvroAnnotationIntrospector extends JacksonAnnotationIntrospector {
+    public static class AvroDeserializer<T> extends JsonDeserializer<T> {
```
cowtowncoder: Why are these inner classes? And what function do they provide (javadoc would be great)

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
```diff
@@ -49,7 +49,17 @@
          *
          * @since 2.7
          */
-        AVRO_BUFFERING(true)
+        AVRO_BUFFERING(true),
+
+        /**
+         * Feature that writes {@code char}, {@code char[]}, {@link Character}, and {@code Character[]} as {@code String strings} instead of
+         * the avro-standard integers.
```
cowtowncoder: Not quite sure what this means....

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
```diff
@@ -28,12 +47,13 @@ public AvroMapper() {
     public AvroMapper(AvroFactory f) {
         super(f);
         registerModule(new AvroModule());
+        setDefaultTyping(new AvroSchemaTypeResolverBuilder());
```
cowtowncoder: I am not sure it is a good idea to force default typing just because Avro encoding is used? Although granted use of Jackson's standard polymorphic typing probably won't work well with most Avro type definitions...

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
baharclerode: I probably will not have much chance to update the PR for the next few days, but will go back and add documentation when I do; These are internal classes that wrap an avro `CustomEncoding` and allow it to be registered into Jackson as a standard `JsonDeserializer`. I'm still working on the counterpart that functions as a `JsonSerializer`. This is to suppor the `@AvroEncode` annotation.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
```diff
@@ -126,4 +146,85 @@ public AvroSchema schemaFrom(File schemaFile) throws IOException
...
+
+        @Override
+        public Object deserializeTypedFromScalar(JsonParser jsonParser, DeserializationContext context) throws IOException {
+            return deserializeTypedFromAny(jsonParser, context);
+        }
+
+        @Override
+        public Object deserializeTypedFromAny(JsonParser jsonParser, DeserializationContext context) throws IOException {
+            // Primitives have no subtypes. Whatever the base type is, is correct!
+            if (!_baseType.isPrimitive() && jsonParser.canReadTypeId()) {
```
cowtowncoder: Typing should never be enabled for primitive types regardless of settings... so former checks seems superfluous.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
```diff
@@ -126,4 +146,85 @@ public AvroSchema schemaFrom(File schemaFile) throws IOException
...
+    public static class AvroSchemaTypeResolverBuilder extends StdTypeResolverBuilder {
+
+        @Override
+        public TypeSerializer buildTypeSerializer(SerializationConfig config, JavaType baseType, Collection<NamedType> subtypes) {
+            // No default typing for Serialization - It's already encoded in the Schema object
+            return null;
+        }
+
+        @Override
+        public TypeDeserializer buildTypeDeserializer(DeserializationConfig config, JavaType baseType, Collection<NamedType> subtypes) {
```
cowtowncoder: I really do not understand what is being attempted here... type id handling should start from `AnnotationIntrospector` that determines if polymorphic type handling is needed. Code here appears to assume that each and every type is polymorphic?

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
```diff
@@ -126,4 +146,85 @@ public AvroSchema schemaFrom(File schemaFile) throws IOException
...
+        @Override
+        public Object deserializeTypedFromAny(JsonParser jsonParser, DeserializationContext context) throws IOException {
+            // Primitives have no subtypes. Whatever the base type is, is correct!
+            if (!_baseType.isPrimitive() && jsonParser.canReadTypeId()) {
+                Object typeId = jsonParser.getTypeId();
+                if (typeId != null) {
+                    try {
+                        return _deserializeWithNativeTypeId(jsonParser, context, typeId);
+                    } catch (InvalidTypeIdException e) {
+                        // If the base type is acceptable, fall through; otherwise, propogate error
```
cowtowncoder: Exceptions are pretty expensive way to signal things, and swallowing them rarely makes sense. Why is this done here?

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroParser.java
```diff
@@ -255,6 +264,16 @@ public JsonLocation getCurrentLocation()
     @Override
     public abstract JsonToken nextToken() throws IOException;
 
+    @Override
+    public boolean canReadTypeId() {
```
cowtowncoder: This method is typically expected to base decision just on format's capability to encode type information...

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/ArrayReader.java
```diff
@@ -18,18 +21,18 @@
 
     protected String _currentName;
     
-    protected ArrayReader(AvroReadContext parent, AvroParserImpl parser)
+    protected ArrayReader(AvroReadContext parent, AvroParserImpl parser, Schema schema)
```
cowtowncoder: I don't want Avro `Schema` type to be exposed at reader level.
Equivalent information may of course.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroReaderFactory.java
```diff
@@ -1,10 +1,27 @@
 package com.fasterxml.jackson.dataformat.avro.deser;
 
-import java.util.*;
-
+import com.fasterxml.jackson.dataformat.avro.deser.ScalarDecoder.BooleanDecoder;
```
cowtowncoder: Please do not change import statements. I use wildcards for 5 or more types (and sometimes earlier).

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
baharclerode: The type deserializer gets called for the `byteValue` field (of type `byte`) in the `PrimitiveInteropTest`. Since in avro schema this is an "int with typeId java.lang.Byte" and Jackson barfs when it tries to process `java.lang.Byte` as a subtype of `byte`, this check seemed necessary. Let me know if there is a different way I should be handling this case.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
baharclerode: That is essentially correct. Avro encodes polymorpic type information in the schema itself (as a `java-class` property). By setting the default typer, I check the schema to see what the actual type should be even when there is no available type information on the POJO for the `AnnotationIntrospector` to return.

If you peek in `InteropTestBase`, you will see that all I'm passing to Jackson as the target type is `Object.class`, a schema, and a byte array. All additional type information comes from the schema, which is how the Apache implementation encodes it.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
baharclerode: This was done to handle the case when Jackson had type information, but the type information in the schema was invalid (for example, if data serialized from type A was being deserialized into type B, which is not a subtype of A or A is not even a class that exists on the class path), then we fall back to using the type Jackson has.

I added the code that queries the subtype resolver after I wrote this, so it could in theory be removed and instead devs would be required to register explicit named subtypes to map these invalid schema type IDs to valid classes on the class path instead, this just seemed like a nicer developer experience when your type names didn't match up 100% for some reason with your POJOs.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroMapper.java
baharclerode: Avro puts all of its type information in the schema and none in the data structure itself, so registering as the default typer seemed to be the easiest way to make Jackson check the schema for type information since there won't be any annotations to indicate type information in situations like deserializing into a `Map<String,Object>`.

cowtowncoder:
Ok. I think that it would actually make sense to divide this into multiple separate PRs.
I don't quite yet follow some of the aspects (most notable idea behind polymorphic type handling; and to some degree Avro's support for `Character` (via `int` I think).

But many of the fixes seem simple enough so that they could be merged more easily, and those would be good candidates for 2.8.

If you don't mind, I can go over lowest hanging fruits first?
I think remaining can mostly just be for `master` anyway.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
baharclerode: It appeared that when Jackson sees a `char[]`, it tries to write it out as an array of strings, and also tries to read it in as an array of strings. Avro instead uses an array of int for handling these, and while I adapted the schema generator, avro generator, and avro parser to handle this properly, I figured someone might want the old behavior still, so I tried to implement an option to restore the old behavior.

baharclerode:
Thanks for the quick response! I will look into rebasing onto 2.8 and see how much work it is, and start cleaning up the code from the feedback you provided.

Probably about half the fixes (Bugs 1, 4, 5, 6, and maybe 7) are low-hanging fruit that could be split out into a separate PR as long as it isn't related to the Type Id use during serialization/parsing. I'll take a look at that on Monday, though it will be painful to split those tests out from the rest of the type-ID related tests.

cowtowncoder:
@baharclerode How about I'll do simplest of 2.8 parts. I'll have some minor changes to structuring of those (just where to add helper methods); I hope it does not cause too many merge problems here.

baharclerode:
Sure, that would be great!

cowtowncoder:
Ok, I think I managed to merge fixes for issues 1, 4, 5, 7 and 8 you listed in description.
I think I also have bit better understanding of type id wrt Avro. But still need to go through code to understand your approach better.
Once again, thank you very much for contributing this -- interoperability is important, and schemas are big part of value of Avro.

Another thought on splitting PR: Adding tests would be great; perhaps so that initially checking in passing tests normally (under expected locations), and failing ones under `failing` (I need to double-check that build skips those, thought I saw some issues earlier). Latter can then be moved to regular packages once fixes are checked in.

cowtowncoder:
... and ok, looks like I managed to introduce a few merge conflicts. Sorry about that. I hope they are easy enough to resolve, just alternate take on routing.

cowtowncoder:
Closing since I merged some of changes, and there's newer PR for the rest.

Issue #15, opened 2016-05-07
[avro] Could not crate .avro file


cowtowncoder:
(moved from https://github.com/FasterXML/jackson-dataformat-avro/issues/42 by @srini-daruna)

I created avro schema from class and got the byte array and want to write the data to avro file. 
How can i do that.?

cowtowncoder:
(copied my original comment from source issue)

For better or worse, Avro dictates different format for files than network transport -- Avro files are supposed to have both prefix and pre-written schema as JSON. Currently this module only does "raw" encoded content, and does not have means to add wrapper stuff.
Addition of header and avro schema would be a good addition I think.

Note, however, than when transferring Avro content over the wire, these are not supposed to be used. I don't know if spark (et al) would even allow use; perhaps they try to auto-detect which mode is used.

cowtowncoder:
Implementation added:

    AvroGenerator.Feature.AVRO_FILE_OUTPUT

enabling of which will then prepend schema as expected by Avro for "file" format (with header, schema and then payload), instead of "raw" (`rfc`) format.

PR #46, opened 2017-02-19
Add Avro file output and test

aosagie:
Hi,

After my team and I wondering (for a while) why we couldn't use avro-tools to open the output of this project, we stumbled upon this issue: #15

This PR addresses it by creating a flag to output in the Avro file format.

I'm new to the internals of this project so I may not have done it in the most optimal way.

Thanks for a decade of building and supporting such a great project.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
@@ -49,7 +49,13 @@ public class AvroGenerator extends GeneratorBase
          *
          * @since 2.7
          */
-        AVRO_BUFFERING(true)
+        AVRO_BUFFERING(true),
+
+        /**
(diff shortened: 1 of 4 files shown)


cowtowncoder:
@aosagie Excellent! Thank you for providing this patch -- it does make sense.
I think it may be possible to slightly optimize this, but I think simplicity of the approach is good as the first step.
So I would be happy to merge this patch, to be included in 2.9.0 (since it requires API add which can't be done in a patch version).

One small process thing we need is to get a Contributor License Agreement (CLA). It only needs to be done, and is good for all further contributions for Jackson projects.
Document itself is at:

(this is individual one; there's corporate one too if need be, but most contributions are under individual one)

and the usual method is to print, fill & sign, scan and email to `info` at fasterxml dot com.
So if you could do this I would be happy to get PR merged ASAP.

Thank you once again.

aosagie:
Sounds good. I've emailed the signed document.

cowtowncoder:
CLA received; will merge.

PR #47, opened 2017-02-20
Avro Array/Character/char[] fixes and Test Suite

baharclerode:
As per the discussion in https://github.com/FasterXML/jackson-dataformats-binary/pull/44, I have rebased the following onto the 2.8 branch:

* Various fixes for Arrays and fixed how Character, Character[], char, and char[] are handled
* Rebuilt the test suite to make it much more fine-grained, so individual tests can be disabled
* Added support for `@AvroIgnore` and `@AvroName`

This should be the rest of the changes in the original PR that did not depend upon the native type ID work, and each commit should be fairly self contained if you want to review commit-by-commit.

I think the only non-trivial merge conflict with master was the `CharReader`, which needs to be updated to look like it does [here](https://github.com/FasterXML/jackson-dataformats-binary/pull/44/files#diff-0156682d65c5d63cbc4d3b03852b5828).

Let me know if there's anything I missed style-wise or if there is a better way to implement one of these features.

### DIFF
--- a/avro/pom.xml
+++ b/avro/pom.xml
@@ -45,6 +45,20 @@ abstractions.
       <version>1.1.3</version>
       <scope>test</scope>
     </dependency>
+    <!-- A bit of help to reduce boiler-plate in dummy test classes -->
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <version>1.16.12</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- For validating more complex comparisons -->
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>3.6.2</version>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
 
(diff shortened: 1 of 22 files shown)


cowtowncoder:
Excellent. With earlier fixes merged earlier this is bit cleaner now, and makes sense to me.
Only one thing I am wondering about: `AnnotationIntrospector` implementation. Instead of extending `JacksonAnnotationIntrospector` I think it should just be stand-alone one, and then it could be combined with others (like JAXB-AI). I can make that change probably myself.

I hope to find time tomorrow to merge and make the minor ai change.

baharclerode:
That sounds perfect. The AI should work just as well stand-alone.

cowtowncoder:
Hmmh. Ugh. This patch accidentally uses Java 8 stuff... need to fix that; we can not use Java 8 before Jackson 3.x.

cowtowncoder:
Hmmh. Must be indirect dep, will try to investigate.

cowtowncoder:
... and, yes, it's AssertJ. Crap. Downgrading to 2.5, as 3.x requires Java 8. Lombok is fine.

PR #48, opened 2017-02-28
[Avro] Include Type IDs in List schemas

baharclerode:
This fixes several issues with type IDs in generated schemas, particularly around lists. The tests suites for maps and lists are also fleshed out, which helped to catch these issues.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
@@ -1,5 +1,7 @@
 package com.fasterxml.jackson.dataformat.avro.schema;
 
+import java.util.List;
+
 import org.apache.avro.Schema;
 import org.apache.avro.Schema.Type;
 
@@ -35,8 +37,8 @@ public class ArrayVisitor
             throw new IllegalStateException("No element schema created for: "+_type);
         }
         Schema schema = Schema.createArray(_elementSchema);
-        if (_type.isArrayType()) {
-            schema.addProp(AVRO_SCHEMA_PROP_CLASS, _type.toCanonical());
+        if (!_type.hasRawClass(List.class)) {
+            schema.addProp(AVRO_SCHEMA_PROP_CLASS, AvroSchemaHelper.getTypeId(_type));
         }
         return schema;
     }
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
@@ -196,6 +196,6 @@ public abstract class AvroSchemaHelper
         if (type.isPrimitive()) {
             return ClassUtil.wrapperType(type.getRawClass()).getName();
         }
-        return type.toCanonical();
+        return type.getRawClass().getName();
     }
 }
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/ApacheAvroInteropUtil.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/ApacheAvroInteropUtil.java
@@ -108,7 +108,7 @@ public class ApacheAvroInteropUtil {
                     }
                 }
             }
-            if (type instanceof Class<?> && ((Class<?>) type).getSuperclass() != null) {
+            if (type instanceof Class<?> && ((Class<?>) type).getSuperclass() != null && !Enum.class.isAssignableFrom((Class<?>) type)) {
                 // Raw class may extend a generic superclass
                 // extract all the type bindings and add them to the map so they can be returned by the next block
                 // Interfaces shouldn't matter here because interfaces can't have fields and avro only looks at fields.
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/InteropTestBase.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/InteropTestBase.java
@@ -1,11 +1,16 @@
 package com.fasterxml.jackson.dataformat.avro.interop;
 
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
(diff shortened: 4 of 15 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
```diff
@@ -196,6 +196,6 @@ protected static String getTypeId(JavaType type) {
         if (type.isPrimitive()) {
             return ClassUtil.wrapperType(type.getRawClass()).getName();
         }
-        return type.toCanonical();
+        return type.getRawClass().getName();
```
baharclerode: `toCanonical` includes generic type information in `<>`, which the Apache implementation cannot understand and will cause exceptions to be thrown.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/ApacheAvroInteropUtil.java
```diff
@@ -108,7 +108,7 @@ protected Schema createSchema(Type type, Map<String, Schema> names) {
                     }
                 }
             }
-            if (type instanceof Class<?> && ((Class<?>) type).getSuperclass() != null) {
+            if (type instanceof Class<?> && ((Class<?>) type).getSuperclass() != null && !Enum.class.isAssignableFrom((Class<?>) type)) {
```
baharclerode: Stop infinite recursion

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
```diff
@@ -35,8 +37,8 @@ public Schema builtAvroSchema() {
             throw new IllegalStateException("No element schema created for: "+_type);
         }
         Schema schema = Schema.createArray(_elementSchema);
-        if (_type.isArrayType()) {
-            schema.addProp(AVRO_SCHEMA_PROP_CLASS, _type.toCanonical());
+        if (!_type.hasRawClass(List.class)) {
+            schema.addProp(AVRO_SCHEMA_PROP_CLASS, AvroSchemaHelper.getTypeId(_type));
```
baharclerode: All list types need type information, not just java arrays

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
cowtowncoder: Ah ok. It would be great to have a test for this, if you have seen such failure?

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/ApacheAvroInteropUtil.java
cowtowncoder: Yes. Atlhough the self-recursive type definitions exist elsewhere too, so same problem could occur with builder-style things. But enums may be the only common example.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
cowtowncoder: Hmmh. Ok, but how about Scala's `List` equivalent? (it doesn't implement `List` but behaves similarly) Does `List` not need it, or is this actively harmful? My general worry is just that exact check for JDK type could be fragile, but I guess it's more due to my not understanding the need for this check.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
baharclerode: The `ListSubtypeTest` will have failing cases when a Jackson schema and Apache deserializer are being used together.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/ApacheAvroInteropUtil.java
baharclerode: True; I'm not super-concerned here though because this is just a local patch to the Apache implementation to make it easier to test in a uniform manner as Jackson; All that matters is that the types we use in the unit tests work.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
baharclerode: The actual behavior required to be compatible with the apache implementation is "If a `GenericData.Array` cannot be assigned to the type, then it must include a type ID" or it won't deserialize properly. This check resulted in slightly cleaner schemas since `List` is a reasonably common field type and doesn't require type information, but we could just do what the Apache implementation does and always include the type information, as it shouldn't hurt anything (the tests will verify this).

baharclerode: And to be a bit more specific, if Scala's `List` equivalent does not at least implement `java.util.Collection`, then it is not compatible with Avro's `ARRAY` type when using the Apache implementation, and would be serialized as a record.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/ApacheAvroInteropUtil.java
cowtowncoder: Ah right. Sorry, failed to notice this is in test helper class so that's fine.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
cowtowncoder: Ok good.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
cowtowncoder: @baharclerode True, but keep in mind that Avro RI encoder wouldn't be seeing original value at all: `AvroGenerator` will essentially re-build contents that were split by `jackson-databind`. So encoder need not be aware of any of that. It will be just passed a `List`. However, type of the POJO property may be something completely different.

I don't have enough context here to know for sure; I guess I am ok with this check for now, and it can be improved upon if and as need be.

Issue #13, opened 2016-05-06
[avro] Add support for Avro default values


cowtowncoder:
(moved from https://github.com/FasterXML/jackson-dataformat-avro/issues/34 by @turbospaces)

Right now RecordVisitor create Fields without default VALUE, which causes problem when the schema evaluates and client operates with OLD data http://ben-tech.blogspot.com/2013/05/avro-schema-evolution.html

I suggest to change Record visitor to:

``` java
@Override
public void optionalProperty(BeanProperty writer) throws JsonMappingException {
  Schema schema = schemaForWriter( writer );
  if ( !writer.getType().isPrimitive() ) {
      schema = AvroSchemaHelper.unionWithNull( schema );
  }
    _fields.add( new Schema.Field( writer.getName(), schema, null,
       writer.isRequired() ? null : Schema.parseJson( "null" ) ) );
}
```

cowtowncoder:
Full support has been added now: this part is covered by full handling of default values during introspection.

PR #49, opened 2017-02-28
[Avro] Add support for @AvroDefault and @Nullable annotations

baharclerode:
This adds support for embedding the default values (Provided by `@AvroDefault` or `@JsonProperty`) into the generated avro schemas. This also adds support for `@Nullable` in case there are multiple annotation introspectors in play and the annotations need to be recognized despite normally being the default.

I seem to have found a weird bug where if a property has a default value or a description, but is neither explicitly marked as required nor not required, `POJOPropertyBuilder.getMetadata():503` will throw an NPE when it tries to coerce the `required` marker into a boolean primitive. I wasn't sure if this was intentional or something I don't fully understand. I worked around it by providing explicit `required` markers if `@AvroDefault` is present, but there's probably a better way to handle this that I'm not aware of. Normally this isn't a problem because `@JsonProperty` doesn't allow you to specify a default value or description without also setting the required marker to `true` or `false`.

Additional PRs for `@AvroSchema`, `@AvroMeta`, and `@Union` will probably be filed tomorrow sometime. The PR for adding polymorphic deserialization backported to the 2.8 branch will follow those.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
@@ -1,5 +1,6 @@
 package com.fasterxml.jackson.dataformat.avro;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.Version;
 import com.fasterxml.jackson.databind.AnnotationIntrospector;
 import com.fasterxml.jackson.databind.PropertyName;
@@ -9,6 +10,7 @@ import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
 import org.apache.avro.reflect.AvroDefault;
 import org.apache.avro.reflect.AvroIgnore;
 import org.apache.avro.reflect.AvroName;
+import org.apache.avro.reflect.Nullable;
 
 /**
  * Adds support for the following annotations from the Apache Avro implementation:
@@ -18,6 +20,7 @@ import org.apache.avro.reflect.AvroName;
  * <li>{@link AvroDefault @AvroDefault("default value")} - Alias for <code>JsonProperty.defaultValue</code>, to
  *     define default value for generated Schemas
  *   </li>
+ * <li>{@link Nullable @Nullable} - Alias for <code>JsonProperty(required = false)</code></li>
  * </ul>
  *
  * @since 2.9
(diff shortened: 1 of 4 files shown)


cowtowncoder:
@baharclerode On NPE, it is fairly safe to say it is not intentional. I can't think of a reason to do so.
I can try to fix that.

Patch makes sense on its own, although I think reordering wrt default value is something that would actually be useful in at least one other place, where converting "resolving" deserializer (that is, one where separate reader/writer schemas are used).

One note on parsing default value: this will be bit of a pain to merge into `master` since I plan on using Avro 1.8 which changes handling to use plain `Object`, instead of Jackson 1.x tree model (to better encapsulate internal details).

baharclerode:
I think my final usecase requires avro 1.8.1 compatibility anyways (and jackson-dataformat-avro 2.8 is _not_ compatible with 1.8.1), so I can just rebase this onto master and reopen a new pull request if you'd like.

cowtowncoder:
I think it'b great if you could add other pieces for Jackson 2.8, and I'll merge those, and then default handling for `master` (2.9)?

cowtowncoder:
Just realized that this will implement/fix #13.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
```diff
@@ -57,4 +60,18 @@ protected PropertyName _findName(Annotated a)
...
+
+    @Override
+    public Boolean hasRequiredMarker(AnnotatedMember m) {
+        if (_hasAnnotation(m, Nullable.class)) {
+            return false;
+        }
+        // Appears to be a bug in POJOPropertyBuilder.getMetadata()
+        // Can't specify a default unless property is known to be required or not, or we get an NPE
+        // If we have a default but no annotations indicating required or not, assume required.
+        if (_hasAnnotation(m, AvroDefault.class) && !_hasAnnotation(m, JsonProperty.class)) {
```
cowtowncoder: I don't think check for `@JsonProperty` should be needed here.
Is that only used to avoid `NPE` or for some other need?

cowtowncoder:
@baharclerode I just fixed what was causing `NPE` you saw, I think. Fix in 2.8 branch, master.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
baharclerode: This is only to work around the NPE. I just saw your message about fixing it, so I'll test everything with that and see if I can remove this workaround!

cowtowncoder:
Btw, not related to this patch but I did some butchering for tests in `master`... apologies. Was having some issues with Lombok (I think), and didn't quite grok why exceptions caught differed (probably since one path wraps exceptions, other not). But on plus side, `master` builds again.

baharclerode:
Building is good! The test changes looked alright at a quick glance. I'm closing out this PR so that I can reopen it against master.

PR #52, opened 2017-03-01
[Avro] Add support for @AvroDefault and @Nullable annotations

baharclerode:
This adds support for embedding the default values (Provided by `@AvroDefault` or `@JsonProperty`) into the generated avro schemas. This also adds support for `@Nullable` in case there are multiple annotation introspectors in play and the annotations need to be recognized despite normally being the default.

(reopened against the `master` branch)


cowtowncoder:
Excellent, thank you!

PR #53, opened 2017-03-01
[Avro] Add support for @Stringable annotation

baharclerode:
This adds support for the `@Stringable` annotation, and adds support for stringable classes (Those with `@Stringable`, and `URL`/`URI`/`File`/`BigInteger`/`BigDecimal`). This fixes BigInteger/BigDecimal compatibility with the apache implementation.

Of particular note, `BigDecimal` is no longer a `Type.DOUBLE` in schema generation, nor is `BigInteger` a `Type.LONG` in schema generation (since they are "stringable" to avro), but they should continue to serialize/deserialize properly from int/long/float/double scalars in existing avro schemas.

I treated `@Stringable` on a class as having a creator annotation on the constructor which takes a single string argument. I'm not sure if there is a better way to do this.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
@@ -1,14 +1,20 @@
 package com.fasterxml.jackson.dataformat.avro;
 
-import com.fasterxml.jackson.core.Version;
-import com.fasterxml.jackson.databind.AnnotationIntrospector;
-import com.fasterxml.jackson.databind.PropertyName;
-import com.fasterxml.jackson.databind.introspect.Annotated;
-import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
+import java.io.File;
 
 import org.apache.avro.reflect.AvroDefault;
 import org.apache.avro.reflect.AvroIgnore;
 import org.apache.avro.reflect.AvroName;
+import org.apache.avro.reflect.Stringable;
+
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.databind.AnnotationIntrospector;
+import com.fasterxml.jackson.databind.PropertyName;
+import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.fasterxml.jackson.databind.introspect.AnnotatedClass;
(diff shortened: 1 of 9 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
```diff
@@ -57,4 +65,23 @@ protected PropertyName _findName(Annotated a)
...
+            a instanceof AnnotatedConstructor && ((AnnotatedConstructor) a).getTypeContext() instanceof AnnotatedClass
+            ? (AnnotatedClass) ((AnnotatedConstructor) a).getTypeContext()
+            : null;
+        return constructor != null && parentClass != null && parentClass.hasAnnotation(Stringable.class)
+            && constructor.getParameterCount() == 1 && String.class.equals(constructor.getRawParameterType(0));
+    }
+
+    @Override
+    public Object findSerializer(Annotated a) {
+        if (a instanceof AnnotatedClass && a.hasAnnotation(Stringable.class) || a.getRawType() == File.class) {
```
cowtowncoder: First part makes sense I guess, but do we really need `File`? It's serialized using `value.getAbsolutePath()` already.
Alternatively if this really needs to be changed, I think we should just register serializer via `AvroModule`, and not through annotation introspector.

cowtowncoder:
Hate to to this but... would it be possible to re-base it to `master`, and only add in 2.9?
There are some aspects that have changed in code regarding annotation introspection (regarding creator detection), and I think I'd like to make some tweaks. I will try to get this in for `2.9.0.pr2`.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
baharclerode: Avro uses `toString()` / `getPath()` instead of `getAbsolutePath()`, so this was a quick fix to make the behavior match up. I'll replace it with a proper serializer in the module.

baharclerode:
I can rebase this into master. After this, I have two more planned PRs, one for `@Union` / type ID support, and one for `@AvroEncode` support.

baharclerode:
Closed and reopened against master as https://github.com/FasterXML/jackson-dataformats-binary/pull/57

Issue #54, opened 2017-03-06
[protobuf] Some fields are left null


niksu7:
Depending on field names and ordering, some fields are at times left out, probably when reading in proto message (since this can be reproduced also by writing the same content with c++ compiled proto interface and reading in with Jackson).

Attached JUnit test case with two ways to work around the problem (JsonPropertyOrder annotation or leaving out a nested object value).

Tested with version 2.8.7 (and same problems in previous versions too, but I haven't run  this exact test with any other version).
[TestProto.java.txt](https://github.com/FasterXML/jackson-dataformats-binary/files/821017/TestProto.java.txt)

niksu7:
Correction: the problem must be in writing part, since the message is only 11 bytes when the nested object (testClass.b.d) is set, but 23 bytes when it is not set.

niksu7:
I have a correction to this now, with minor changes to ProtobufGenerator.java and ByteAccumulator.java.

cowtowncoder:
Thank you for the patch & test! I will backport this in 2.8.

cowtowncoder:
As per notes on PR itself, it would appear that fix does not fix test failure (although naming of test class prevents it from getting run via Maven, hiding the fail).

niksu7:
Oops, I'll check that today. I ran the test manually from Eclipse while testing the fix. It might be that I didn't test the latter manually as I assumed Maven would run the test...

niksu7:
Yes, there was a stupid bug in the last commit.
Also renamed test + refactored it a bit.

cowtowncoder:
Perfect, thank you! I merged it manually for 2.8 so will close PR but changes are the same.
Thank you once again for fixing this. Will be in 2.8.8 and 2.9.0.pr2

cowtowncoder:
Hmmh. I suspect this is related to (or causing) #67, fwtw. Reordering of buffered content, perhaps harmless in some cases.

niksu7:
I'll check. But if the test fails in 2.8.7, it can't be caused by this
change as it's not in 2.8.7.

On 5 Apr 2017 01:25, "Tatu Saloranta" <notifications@github.com> wrote:

Hmmh. I suspect this is related to (or causing) #67
<https://github.com/FasterXML/jackson-dataformats-binary/issues/67>, fwtw.
Reordering of buffered content, perhaps harmless in some cases.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
[...]

PR #55, opened 2017-03-07
Fix issue #54: Some fields are left null

niksu7:
Patch for fixing #54


cowtowncoder:
Hmmh. I can't replicate fix -- it seems that test class is named `NestedMessages` and as such does not get run (junit tests use `Test*` and `*Test` matching) as is. But when renaming, test appears to fail unless I did something wrong.

cowtowncoder:
Merge manually in 2.8 so closing PR.

PR #57, opened 2017-03-09
[Avro] Add support for @Stringable annotation

baharclerode:
This adds support for the `@Stringable` annotation, and adds support for stringable classes (Those with `@Stringable`, and `URL`/`URI`/`File`/`BigInteger`/`BigDecimal`). This fixes BigInteger/BigDecimal compatibility with the apache implementation.

Of particular note, `BigDecimal` is no longer a `Type.DOUBLE` in schema generation, nor is `BigInteger` a `Type.LONG` in schema generation (since they are "stringable" to avro), but they should continue to serialize/deserialize properly from int/long/float/double scalars in existing avro schemas.

I treated `@Stringable` on a class as having a creator annotation on the constructor which takes a single string argument. I'm not sure if there is a better way to do this.

----

Rebased onto master as per [this comment](https://github.com/FasterXML/jackson-dataformats-binary/pull/53#issuecomment-285263097).

### DIFF
--- a/avro/pom.xml
+++ b/avro/pom.xml
@@ -21,7 +21,12 @@ abstractions.
   </properties>
 
   <dependencies>
-    <!--  Hmmh. Need databind for Avro Schema generation... -->
+    <!--  Hmmh. Need annotations for introspection... -->
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+    </dependency>
+    <!--  and databind for Avro Schema generation... -->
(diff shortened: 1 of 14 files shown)


cowtowncoder:
Hmmh. This change seems to have nasty performance downside: serialization 30% slower, deserialization 5-15%...

cowtowncoder:
Ok, was able to make minor changes to get serialization perf back: I do not think `@Stringable` check ought to be needed on actual write path (since annotation lookups are remarkable ineffecient). May do more tweaks, but works fine for now.

PR #65, opened 2017-03-21
Issue with broken surrogate pairs

IanWoods1993:
The test in the class BrokenSurrogatePair demonstrates an issue that appears to be rooted somewhere in the SMILE format. The byte array at the top includes a byte sequence that is outside the BMP (https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane). There is no problem writing the value to a JSON node, but when trying to write it in SMILE format, the program crashes:
`com.fasterxml.jackson.databind.JsonMappingException: Broken surrogate pair: first char 0xd92f, second 0x20; illegal combination`

This branch isn't intended to fix this problem, just demonstrate its existence in the hopes that it can be more easily identified and fixed. Thanks!

### DIFF
--- a/jackson-bom-pom.xml
+++ b/jackson-bom-pom.xml
@@ -1,0 +1,288 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.fasterxml.jackson</groupId>
+        <artifactId>jackson-parent</artifactId>
+        <version>2.9.pr1b</version>
+    </parent>
+
+    <artifactId>jackson-bom</artifactId>
+    <version>2.9.0.pr2-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <url>https://github.com/FasterXML/jackson-bom</url>
+    <scm>
+        <connection>scm:git:git@github.com:FasterXML/jackson-bom.git</connection>
+        <developerConnection>scm:git:git@github.com:FasterXML/jackson-bom.git</developerConnection>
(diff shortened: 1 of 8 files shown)


jackson-bom-pom.xml
```diff
@@ -0,0 +1,288 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cowtowncoder: I suspect this got added accidentally. Since there is `jackson-bom` I don't think this is needed.

smile/src/main/java/com/fasterxml/jackson/dataformat/smile/SmileFactory.java
```diff
@@ -7,6 +7,7 @@
 import com.fasterxml.jackson.core.format.InputAccessor;
 import com.fasterxml.jackson.core.format.MatchStrength;
 import com.fasterxml.jackson.core.io.IOContext;
+import com.fasterxml.jackson.core.json.PackageVersion;
```
cowtowncoder: No, this is wrong. Please remove. `PackageVersion` gets generated by Maven build, and we must use one that is local here and not one from another package.
(I understand why this can occur when IDE complains, if no Maven build precedes etc).

cowtowncoder:
@IanWoods1993 Ah! Ok, ignore notes in patch part then -- test case is what matters. Gotcha.

As to problem itself, I'll have a look to make sure I understand.
Thank you for reporthing this.

cowtowncoder:
Hmmh. Content you are trying to write is illegal Unicode: there is an unmatched Surrogate Pair, which is not legal for UTF-8 encoding. So I think exception makes sense here.

It is odd that this did not trigger decoding error on JSON side, however; I will see why this happens.

IanWoods1993:
Thanks so much for looking at this so quickly. Really appreciate it.

cowtowncoder:
@IanWoods1993 No prob, problems wrt low-level encoding/decoding are high priority for me.
I will create a `jackson-core` issue as a follow-up here.
It is also possible that perhaps read/write features to allow broken surrogate use would make sense (that is, to write illegal surrogate pairs, or decode ones) -- this mostly because Java `String`s and `char`s do allow creation of such combinations.

PR #69, opened 2017-03-28
[Avro] Add support for @AvroEncode annotation

baharclerode:
This enables support for the `@AvroEncode` annotation, which is the avro equivalent of a `JsonSerializer`/`JsonDeserializer`.

The serialization side of things was fairly straight-forward, since we can write directly to the underlying datum without worrying about parser state; Deserialization is a bit more tricky, since the parser inserts virtual tokens (start/end object, field names on records, etc.) that we need to consume those and otherwise ensure the parser state is updated so that deserialization control can be handed back to Jackson once the `CustomEncoding` has done its job.

This is the last major piece of the puzzle that brings compatibility with the apache implementation up to an acceptable level for my use cases, so hoping it'll make the cut for 2.9 😃 

----

I also re-enabled the tests that were ignored in 2.8 due to namespace changes, since the underlying code changes were themselves reverted when merged into master a while back.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
@@ -20,7 +20,10 @@ import com.fasterxml.jackson.databind.introspect.AnnotatedMember;
 import com.fasterxml.jackson.databind.jsontype.NamedType;
 import com.fasterxml.jackson.databind.jsontype.TypeResolverBuilder;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import com.fasterxml.jackson.databind.util.ClassUtil;
+import com.fasterxml.jackson.dataformat.avro.deser.CustomEncodingDeserializer;
 import com.fasterxml.jackson.dataformat.avro.schema.AvroSchemaHelper;
+import com.fasterxml.jackson.dataformat.avro.ser.CustomEncodingSerializer;
 
 /**
  * Adds support for the following annotations from the Apache Avro implementation:
@@ -60,6 +63,15 @@ public class AvroAnnotationIntrospector extends AnnotationIntrospector
     @Override
     public PropertyName findNameForDeserialization(Annotated a) {
         return _findName(a);
+    }
+
+    @Override
+    public Object findDeserializer(Annotated am) {
+        AvroEncode ann = _findAnnotation(am, AvroEncode.class);
+        if (ann != null) {
+            return new CustomEncodingDeserializer<>((CustomEncoding)ClassUtil.createInstance(ann.using(), true));
+        }
+        return null;
     }
 
     @Override
@@ -112,6 +124,10 @@ public class AvroAnnotationIntrospector extends AnnotationIntrospector
     public Object findSerializer(Annotated a) {
         if (a.hasAnnotation(Stringable.class)) {
             return ToStringSerializer.class;
+        }
+        AvroEncode ann = _findAnnotation(a, AvroEncode.class);
+        if (ann != null) {
+            return new CustomEncodingSerializer<>((CustomEncoding)ClassUtil.createInstance(ann.using(), true));
         }
         return null;
     }
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
@@ -8,18 +8,11 @@ import java.nio.ByteBuffer;
 
 import org.apache.avro.io.BinaryEncoder;
 
-import com.fasterxml.jackson.core.Base64Variant;
-import com.fasterxml.jackson.core.FormatFeature;
-import com.fasterxml.jackson.core.FormatSchema;
-import com.fasterxml.jackson.core.JsonGenerationException;
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.core.ObjectCodec;
-import com.fasterxml.jackson.core.PrettyPrinter;
-import com.fasterxml.jackson.core.SerializableString;
-import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.base.GeneratorBase;
 import com.fasterxml.jackson.core.io.IOContext;
 import com.fasterxml.jackson.dataformat.avro.ser.AvroWriteContext;
+import com.fasterxml.jackson.dataformat.avro.ser.EncodedDatum;
 
 public class AvroGenerator extends GeneratorBase
 {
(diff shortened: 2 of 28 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
```diff
@@ -63,6 +65,15 @@ public PropertyName findNameForDeserialization(Annotated a) {
     }
 
     @Override
+    public Object findDeserializer(Annotated am) {
+        AvroEncode ann = _findAnnotation(am, AvroEncode.class);
+        if (ann != null) {
+            return CustomEncodingDeserializer.class;
```
baharclerode: `AnnotationIntrospectorPair` does not support returning `JsonDeserializer` _instances_, so using a `HandlerInstantiator` to instantiate the deserializer with the correct context information.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
```diff
@@ -113,6 +124,10 @@ public Object findSerializer(Annotated a) {
         if (a.hasAnnotation(Stringable.class)) {
             return ToStringSerializer.class;
         }
+        AvroEncode ann = _findAnnotation(a, AvroEncode.class);
+        if (ann != null) {
+            return CustomEncodingSerializer.class;
```
baharclerode: `AnnotationIntrospectorPair` does not support returning `JsonSerializer` _instances_, so using a `HandlerInstantiator` to instantiate the serializer with the correct context information.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
cowtowncoder: Hmmh. What do you mean by not supporting that? Since its `Object` it should, and if not, that should be fixed...

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
cowtowncoder: Ditto here, my reading would suggest instances or classes both should work?
(but I guess if there's problem with one there's with the other too)

cowtowncoder: Or do you mean that since there isn't enough information, it may not be done? Contextualization will be done, however, so as long as serializer/deserializer implement `Contextual(De)Serializer`, matching `createContextual()` method will be called to give access to annotations, active configuration.

cowtowncoder:
Ok hmmh. This is quite elaborate, and I'll have to think carefully... since the problem is that I would ideally want to reduce or eventually even eliminate dependencies to the standard avro lib, wrt encoding/decoding. If this is only sort of isolated area and functionality only affects annotated types (which I think is the case), it may still be ok. And I am +1 on increased interoperability per se.
I just don't want to add coupling to implementation; for example, to improve serialization speed I would eventually want to implement encoding directly.

I'll have to read over the code a bit more; I like many of the aspects (and we can probalby simplify some parts as per comments I added).

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
baharclerode: The `_isExplicitClassOrOb` check [here](https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/introspect/AnnotationIntrospectorPair.java#L315) requires that the primary return a `Class` object - If you return an instance of a `JsonSerializer`, `_isExplicitClassOrOb` returns false and the secondary is used. If that's unintended behavior, then yeah, that should be fixed, and this code could be simplified to just return a `JsonSerializer` instance with the context information already bound instead of deferring that to a custom `HandlerInstantiator` (and the same for `JsonDeserializer` above)

baharclerode:
You won't be able to completely remove the encoding dependencies and support this annotation - The annotation specifically requires a subtype of `CustomEncoding`, which references both the `Decoder` and `Encoder` interfaces; They are, however, just interfaces, which means that it will still be completely decoupled from the implementation.

For example, the deserialization should already be decoupled, since that is handled by a wrapper that wraps the `AvroParserImpl` and implements `Decoder` using the existing methods to ensure that the parser state is advanced correctly. I don't think it even calls any of the decoder primitives that `AvroParserImpl` exposes.

Serialization would probably require a similar pattern, but as long as the underlying implementation supports the same primitives, the `NonBSGenericDatumWriter` can be swapped out for any other (faster) implementation that knows how to pass itself or a wrapper into `EncodedDatum.write()` method.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
baharclerode: I think changing [this](https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/introspect/AnnotationIntrospectorPair.java#L788) to

    return true;

would fix the issue, and would actually make the `OrOb` part of the method name make more sense, now that I think about it.

cowtowncoder: Ok intention was to be reflected by method name: "Is (explicit) Class or Object". The only (?) kink there was simply that there are placeholder values to indicate "This is not a real value" -- just because Annotations themselves can NOT have value of `null`, ever, so if a Class-value annotation optional property is desired, it MUST have value of some bogus class.

So, long way of saying: yes, returning a (de)serializer absolutely should be supported and if not, there's a bug. I'll file a bug on `jackson-databind` and test this, for 2.9.

cowtowncoder:
Agreed: I am ok with using annotations as well as simple interfaces (Decoder, Encoder) for this part.
I am only worried about keeping those isolated, so that not everything has to implement these abstractions.
And if at some future point it seemed useful we could further modularize things to isolate even these aspects -- but that might never happen.

But note that my explicit goal (hoping to still get done for 2.9) is to remove actual usage of `BinaryDecoder` for standard decoding. I don't think this would interfere with that goal, I just want to mention it just in case I am overlooking something.

baharclerode:
Your fix for `_isExplicitClassOrOb` worked perfectly, and I've updated the PR to remove the `AvroHandlerInstantiator`.

cowtowncoder:
@baharclerode Looks like there was yet another smaller issue, now fixed too:

https://github.com/FasterXML/jackson-databind/issues/1584

but I don't think it'd affect your changes here.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
cowtowncoder: @baharclerode Ok, filed & fixed https://github.com/FasterXML/jackson-databind/issues/1581 -- it was regression from 2.8, but wasn't covered by unit tests (ironically enough I had added some tests for `AnnotationIntrospectorPair` in 2.9... but not for this method).

PR #71, opened 2017-04-04
[Avro] Fix custom deserializers not being used

baharclerode:
Found an issue that wasn't caught by one of the test suites.

Turns out because I'm always returning a `TypeDeserializer` for every property, Jackson doesn't use any custom deserializers that are registered for that property. I have a workaround here that checks for one and uses it instead, but I think there's a better way whereby I don't return a `TypeResolverBuilder` for properties that have custom deserializers, but I couldn't get that working. What's the proper way to fix this?

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroTypeDeserializer.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroTypeDeserializer.java
@@ -4,10 +4,7 @@ import java.io.IOException;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.databind.BeanProperty;
-import com.fasterxml.jackson.databind.DeserializationContext;
-import com.fasterxml.jackson.databind.JavaType;
-import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.*;
 import com.fasterxml.jackson.databind.jsontype.TypeDeserializer;
 import com.fasterxml.jackson.databind.jsontype.TypeIdResolver;
 import com.fasterxml.jackson.databind.jsontype.impl.TypeDeserializerBase;
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/AvroEncodeTest.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/AvroEncodeTest.java
@@ -1,11 +1,10 @@
 package com.fasterxml.jackson.dataformat.avro.interop.annotations;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
+import java.nio.ByteBuffer;
+import java.util.*;
 
+import org.apache.avro.SchemaBuilder;
 import org.apache.avro.io.Decoder;
 import org.apache.avro.io.Encoder;
 import org.apache.avro.reflect.AvroEncode;
@@ -53,6 +52,9 @@ public class AvroEncodeTest extends InteropTestBase {
         @Nullable
         public Long longValue;
 
+        @AvroEncode(using = UuidAsBytesAvroEncoding.class)
+        private UUID uuidValue;
+
         protected CustomComponent() { }
     }
 
@@ -60,7 +62,7 @@ public class AvroEncodeTest extends InteropTestBase {
     public static class ApacheImplEncoding extends CustomEncoding<CustomComponent> {
 
         public ApacheImplEncoding() throws IOException {
-            schema = ApacheAvroInteropUtil.getJacksonSchema(CustomComponent.class);
+            schema = ApacheAvroInteropUtil.getApacheSchema(CustomComponent.class);
         }
 
         @Override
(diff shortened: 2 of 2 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroTypeDeserializer.java
```diff
@@ -57,6 +54,13 @@ public Object deserializeTypedFromScalar(JsonParser p, DeserializationContext ct
 
     @Override
     public Object deserializeTypedFromAny(JsonParser p, DeserializationContext ctxt) throws IOException {
+        AnnotationIntrospector ai = ctxt.getAnnotationIntrospector();
```
baharclerode: I feel like instead of having the type deserializer look for custom deserializers, there should be some way to avoid using the type deserializer at all for properties that have custom deserializers - but I don't know where that check needs to be made.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroParserImplDecoder.java
```diff
@@ -138,6 +138,13 @@ public void skipString() throws IOException {
     @Override
     public ByteBuffer readBytes(ByteBuffer old) throws IOException {
         consumeToken(JsonToken.VALUE_EMBEDDED_OBJECT);
+        byte[] value = _parser.getBinaryValue();
```
baharclerode: Bugfix: Caller expecting `old` to be populated and not actually using the return value needs to be supported.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/AvroEncodeTest.java
```diff
@@ -74,8 +76,67 @@ protected CustomComponent read(Object reuse, Decoder in) throws IOException {
 
     }
 
-    protected Wrapper wrapper;
+    public static class UuidAsBytesAvroEncoding extends CustomEncoding<UUID> {
```
baharclerode: The original format of this test meant that if the custom deserializer wasn't picked up, Jackson was still flexible enough that it would pass the test because it could read what it wrote. This `UuidAsBytesAvroEncoding` actually changes the serialized format, which means this will now break if the custom serializers/deserializers are ignored.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroTypeDeserializer.java
cowtowncoder: The whole idea of type (de)serializers is that they can be composed with content (de)serializers and work independently. But use of type (de)serializers is inferred from polymorphic handling, so perhaps that should not be enabled if no type handling is desired?

cowtowncoder:
Ouch. Type (de)serializers absolute and positively should never be used for non-polymorphic properties, so that should not be done. I guess I did not read the patch carefully enough.
I am also bit worried about performance aspects, but first problem is of course correctness.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroTypeDeserializer.java
```diff
@@ -57,6 +54,13 @@ public Object deserializeTypedFromScalar(JsonParser p, DeserializationContext ct
...
+        if (ai != null && _property != null) {
+            JsonDeserializer<?> deser = ctxt.deserializerInstance(_property.getMember(), ai.findDeserializer(_property.getMember()));
```
cowtowncoder: This is extremely inefficient as well; usually deserializers are cached, and if this is not done much (most?) of the time is spent on fetching and contextualizing deserializer...

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroParserImplDecoder.java
```diff
@@ -138,6 +138,13 @@ public void skipString() throws IOException {
...
+        if (old.capacity() >= value.length) {
+            old.clear();
+            old.put(value);
+            old.flip();
+            return old;
+        }
         return ByteBuffer.wrap(_parser.getBinaryValue());
```
cowtowncoder: This should pass `value`, and not call accessor again.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroTypeDeserializer.java
baharclerode: Yeah, this was more in the "it _worked_" category rather than the "this is how to fix it" category.

I think I found a better solution, about to push up some changes.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroAnnotationIntrospector.java
```diff
@@ -161,6 +162,10 @@ public Object findSerializer(Annotated a) {
     }
 
     protected TypeResolverBuilder<?> _findTypeResolver(MapperConfig<?> config, Annotated ann, JavaType baseType) {
+        // If there's a custom deserializer, then it's responsible for handling any type information
+        if (config.getAnnotationIntrospector().findDeserializer(ann) != null) {
```
baharclerode: Is this the best way to do this check?

cowtowncoder: I don' t think this approach makes sense: annotations are just one way to register deserializers; modules register others. But beyond this, as I said, type handlers and content handlers are not meant to be aware of each other.

cowtowncoder:
Ok, so, yeah. I don't think `AnnotationIntrospector` path works: there should not be `TypeDeserializer` for things that do not need polymorphic handling.
What might work for some cases is to look for `@Union`, since its existence does signal need for unions. But since schema information isn't otherwise available, I am not sure this part of Avro's type system can actually be support by Jackson currently.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroParserImplDecoder.java
```diff
@@ -138,7 +138,14 @@ public void skipString() throws IOException {
     @Override
     public ByteBuffer readBytes(ByteBuffer old) throws IOException {
         consumeToken(JsonToken.VALUE_EMBEDDED_OBJECT);
-        return ByteBuffer.wrap(_parser.getBinaryValue());
+        byte[] value = _parser.getBinaryValue();
```
cowtowncoder: How about I merge this separate from PR as it seems legit separate from other question.

baharclerode:
Following the discussion on #60, the recent changes to master have fixed this problem. I have updated this PR just to include the tests to catch the issue if it appears again.

cowtowncoder:
Makes sense, merged.

Issue #76, opened 2017-04-11
[protobuf] can't deserialize packed repeated field


knoguchi:
Packed repeated field failed to deserialize the Protobuf bytes.
```
message t1 {
        repeated uint32 report_code = 1 [packed=true];
}
```

The error:

```
com.fasterxml.jackson.core.JsonParseException: Incompatible wire type (0x2) for field 'report_code': not valid for field of type VINT32_STD (expected 0x0)

	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1702)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:558)
	at com.fasterxml.jackson.dataformat.protobuf.ProtobufParser._reportErrorF(ProtobufParser.java:2455)
	at com.fasterxml.jackson.dataformat.protobuf.ProtobufParser._reportIncompatibleType(ProtobufParser.java:2461)
	at com.fasterxml.jackson.dataformat.protobuf.ProtobufParser.nextFieldName(ProtobufParser.java:1110)
	at com.fasterxml.jackson.databind.deser.std.BaseNodeDeserializer.deserializeObject(JsonNodeDeserializer.java:210)
	at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:69)
	at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:15)
	at com.fasterxml.jackson.databind.ObjectReader._bindAndClose(ObjectReader.java:1626)
	at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:1237)
```

When the packed is true, the wire type 0x2 is actually correct as the official document explained in the [Message Structure wire type](https://developers.google.com/protocol-buffers/docs/encoding) section.

Here is the [test code](https://gist.github.com/knoguchi/7a94a8aaec5902d34a6ef1d93e0b31b9). `PackedWrapper` class was generated by `protoc`.

cowtowncoder:
Thank you for submitting this, and fix too. I'll have a look because one thing I don't understand is why existing tests didn't catch this -- packed types are tested, but test coverage is not as extensive as it should.

cowtowncoder:
[...]

PR #77, opened 2017-04-11
Issue #76: fix wire type checking for packed repeated fields

knoguchi:
Fixed the isValidFor method.  The deserializing of the repeated field to an array seems to be fine.

### DIFF
--- a/protobuf/src/main/java/com/fasterxml/jackson/dataformat/protobuf/schema/ProtobufField.java
+++ b/protobuf/src/main/java/com/fasterxml/jackson/dataformat/protobuf/schema/ProtobufField.java
@@ -191,7 +191,7 @@ public class ProtobufField
     }
 
     public final boolean isValidFor(int typeTag) {
-        return (typeTag == type.getWireType());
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Sounds reasonable, but would it be possible to have a test case? This would help against regression.

knoguchi:
Added the test case.   Currently Travis fails for the master branch.   An Avro test case is failing but that's nothing to do with my change.

cowtowncoder:
Ok I had accidentally added failing avro test under other tests, not under `failing`; moved so it won't block builds.

cowtowncoder:
Ok I'm happy to merge this fix (and related) test. Just one thing before merging the contribution; if we haven't yet asked, we'd need a Contributor License Agreement (CLA):

and the easiest way is often to just print, fill & sign, scan, email to `info` at fasterxml dot com.
If that could be done I'd be happy to merge this and then tackle encoding side.

Thank you in advance!

knoguchi:
I've signed and sent the CLA as instructed.    Thank you.

cowtowncoder:
@knoguchi For some reason I don't seem to have that... will try to check my inbox at fasterxml (I get `info` forward to my account).

PR #81, opened 2017-04-19
[Avro] Fix nested record field names returning wrong type ID when deserialized as Object

baharclerode:
If you have nested objects:

```java
public class Level1 {
    public Level2 inner;
}

public class Level2 {
    public Level3 inner;
}

public class Level3 {
    public String value;
}
```

and the types cannot be found during deserialization, it will fail because it gets mapped to `Map<Object, Object>` which triggers polymorphic type lookups on keys, and the `RecordReader` does not report type ID of field names correctly:

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: -1
	at com.fasterxml.jackson.dataformat.avro.deser.RecordReader.getTypeId(RecordReader.java:74)
	at com.fasterxml.jackson.dataformat.avro.deser.RecordReader$Std.getTypeId(RecordReader.java:83)
	at com.fasterxml.jackson.dataformat.avro.deser.AvroFieldReader$Structured.getTypeId(AvroFieldReader.java:66)
	at com.fasterxml.jackson.dataformat.avro.deser.RecordReader.getTypeId(RecordReader.java:71)
	at com.fasterxml.jackson.dataformat.avro.deser.RecordReader$Std.getTypeId(RecordReader.java:83)
	at com.fasterxml.jackson.dataformat.avro.AvroParser.getTypeId(AvroParser.java:240)
	at com.fasterxml.jackson.dataformat.avro.AvroUntypedDeserializer.mapObject(AvroUntypedDeserializer.java:76)
	at com.fasterxml.jackson.databind.deser.std.UntypedObjectDeserializer.deserialize(UntypedObjectDeserializer.java:220)
	at com.fasterxml.jackson.dataformat.avro.AvroUntypedDeserializer.deserialize(AvroUntypedDeserializer.java:35)
	at com.fasterxml.jackson.databind.jsontype.impl.TypeDeserializerBase._deserializeWithNativeTypeId(TypeDeserializerBase.java:262)
	at com.fasterxml.jackson.dataformat.avro.AvroTypeDeserializer.deserializeTypedFromAny(AvroTypeDeserializer.java:67)
	at com.fasterxml.jackson.databind.deser.std.UntypedObjectDeserializer.deserializeWithType(UntypedObjectDeserializer.java:283)
	at com.fasterxml.jackson.dataformat.avro.AvroUntypedDeserializer.deserializeWithType(AvroUntypedDeserializer.java:48)
	at com.fasterxml.jackson.databind.deser.impl.TypeWrappedDeserializer.deserialize(TypeWrappedDeserializer.java:68)
	at com.fasterxml.jackson.databind.ObjectReader._bindAndClose(ObjectReader.java:1612)
	at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:1245)
```
(precise error arises because the type ID of the field value is read before the start token is consumed for that object)

This PR adds a test for this and fixes the issue.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/RecordReader.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/RecordReader.java
@@ -3,6 +3,7 @@ package com.fasterxml.jackson.dataformat.avro.deser;
 import java.io.IOException;
 
 import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.dataformat.avro.schema.AvroSchemaHelper;
 
 abstract class RecordReader extends AvroStructureReader
 {
@@ -68,9 +69,9 @@ abstract class RecordReader extends AvroStructureReader
             return super.getTypeId();
         }
         if (_currToken == JsonToken.FIELD_NAME) {
-            return _fieldReaders[_index].getTypeId();
+            return AvroSchemaHelper.getTypeId(String.class);
         }
-        // When reading a value type ID, the index pointer has already advanced to the field, so look at the previous
+        // When reading a value type ID, the index pointer has already advanced to the next field, so look at the previous
         return _fieldReaders[_index - 1].getTypeId();
     }
 
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/InteropTestBase.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/InteropTestBase.java
@@ -35,7 +35,7 @@ public abstract class InteropTestBase
      *
      * @return A type representing the bound {@code baseClass}
      */
-    protected static ParameterizedType type(Class<?> baseClass, Type... parameters) {
+    public static ParameterizedType type(Class<?> baseClass, Type... parameters) {
         if (baseClass.getTypeParameters().length != parameters.length) {
             throw new IllegalArgumentException("Incorrect number of type parameters, expected "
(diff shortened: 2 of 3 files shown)


cowtowncoder:
Ok not sure I understand why type id would be accessed for key in the first place, but I trust that returning type of value is wrong.

baharclerode:
When deserializing into `Map<Object,Object>`, it checks type ID to find the key type. This is how [non-string keys](https://github.com/FasterXML/jackson-dataformats-binary/blob/master/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/StringableTest.java#L123) are supported.

cowtowncoder:
@baharclerode  Hmmh. Ok, thank you for explanation. It is format-specific extension then.

Btw, not related to this change at all, but maybe worth mentioning: I am completing "native" decoder implementation, and plan to add it for 2.9.0.pr3. Should be fine, passes all tests and so on. Decoding is bit faster, although not drastically so (15-20% in my testing, but depends on kind of content).
But I will leave "classic" variant that uses apache avro codec, so that we can try to verify compatibility easily, and as a fallback in case bugs are uncovered with newer codec.
Configurability will probably be by just choosing sub-class of `AvroFactory` with different codec; not planning anything fancier.

baharclerode:
@cowtowncoder I think that makes sense. It should be relatively easy to add additional combinations to the test suite such that both the classic codec and the newer codec are both tested to ensure compatibility.

cowtowncoder:
@baharclerode yes exactly was thinking that too.

PR #83, opened 2017-04-26
[Avro] Use configured introspector in module

baharclerode:
The module was ignoring the configured introspector (set via `withAnnotationIntrospector`)


cowtowncoder:
Sigh. Knew I should have added tests. Thanks!

Issue #86, opened 2017-05-05
[protobuf] support import statement


knoguchi:
proto files often import other proto files using the `import` statement.   As far as I tested the import statement is not supported.

I think `NativeProtobufSchema` class is the best place to load/cache the external proto dependecy, and resolve message types if they are not found in the `_nativeTypes`.

Do you think it's a sane idea?

cowtowncoder:
Importing is a big challenge: it is important for larger schemas, but is tricky to handle and perhaps potential security concern (at least in XML world ability to reference to DTD content via entities has caused problems -- not exactly identical, but often allows access to files that couldn't accessed otherwise). This may or may not be a big problem since calling code still needs to request reading of schema, and this is not done automatically for content... but something to consider.

So I think that yes, allowing imports would be good, but probably needs to be configurable, to allow pluggable resolution. This could be used to implement caching, but without requiring this module to provide it.

Issue #84, opened 2017-04-29
[protobuf] can't parse `reserved` field


knoguchi:
This is a Protoparser problem that it failed to parse `reserved` field.

Reserved field is explained in the official document.
https://developers.google.com/protocol-buffers/docs/proto

```
va.lang.IllegalStateException: Syntax error in Unnamed-protobuf-schema at 389:11: unexpected label: reserved

	at com.squareup.protoparser.ProtoParser.unexpected(ProtoParser.java:948)
	at com.squareup.protoparser.ProtoParser.readDeclaration(ProtoParser.java:197)
	at com.squareup.protoparser.ProtoParser.readMessage(ProtoParser.java:219)
	at com.squareup.protoparser.ProtoParser.readDeclaration(ProtoParser.java:153)
	at com.squareup.protoparser.ProtoParser.readProtoFile(ProtoParser.java:92)
	at com.squareup.protoparser.ProtoParser.parse(ProtoParser.java:61)
	at com.fasterxml.jackson.dataformat.protobuf.schema.ProtobufSchemaLoader._loadNative(ProtobufSchemaLoader.java:167)
	at com.fasterxml.jackson.dataformat.protobuf.schema.ProtobufSchemaLoader._loadNative(ProtobufSchemaLoader.java:161)
	at com.fasterxml.jackson.dataformat.protobuf.schema.ProtobufSchemaLoader.loadNative(ProtobufSchemaLoader.java:135)
	at com.fasterxml.jackson.dataformat.protobuf.schema.ProtobufSchemaLoader.load(ProtobufSchemaLoader.java:80)
```

cowtowncoder:
Do you know if this is a new addition to particular version of protobuf specification?

knoguchi:
I did a bit of research.   It seems `reserved` was added when protobuf v3.0.0-beta was released 2 years ago for both proto2 and proto3 syntax.

The protoc-2.6.1 can't parse the syntax.  So it's only available in the protobuf v3.x.x releases.  It's a bit confusing but the release version and syntax version is not necessarily corresponding . The proto2 syntax is maintained in the v3 releases.

----

For now I just commented out the `reserved` fields.  So this is not a big issue.

cowtowncoder:
Ok thank you for checking this out. I was just trying to figure this out since I did not remember seeing this keyword before myself.
It seems like a useful thing to have, of course, even with proto v2 as it does not change encoding format. And it is useful for general schema management.

Issue #78, opened 2017-04-12
[protobuf] create ProtoSchema from descriptor file


knoguchi:
I would like to create the ProtoSchema object from the descriptor file instead of proto file.    `protoc -o output.desc` can generate a single descriptor file `output.desc` by compiling multiple proto files.  It's often easier to use the generator file than dealing with multiple .proto files.  At my work we have fairly complex proto files.

I actually have an [experimental implementation](https://gist.github.com/knoguchi/211f44a2137b5def37fbdc04a6db437e).  I used Google's official protobuf-java and [protobuf-dynamic](https://github.com/os72/protobuf-dynamic) to read the descriptor file. 

My questions

1.   would you consider adding the ProtobufSchema by descriptor file feature?
2.  if the answer is yes, do you think the above approach is acceptable i.e. use google's official protobuf jars, and this._source being null.
3.  If the answer is no, would you add a public constructor to the ProtobufSchema so that the 3rd party library can create ProtobufSchema objects without ProtoFile object.

thanks.

cowtowncoder:
I'll have to think about this a bit. But let's start with easy answers....

First of all, yes, allowing use of descriptor files sounds like a good addition, so +1 for that.
I like the idea a lot, since as you say, dealing with multiple complex proto files gets tricky.

As to adding new dependencies: I am bit on the fence on that one -- I prefer minimizing dependencies. But there is no strict limit on what can be added.

So perhaps the best way would be to first start with (3), as I would be happy to add extension points first, to make this possible. And then if and when things work nicely, see if some or all of functionality could be added here.
Or maybe it could be an extension module as sub-project here -- that might be the best way? That is, another maven sub-module (`jackson-module-protobuf-schema` or something), which would be built and released along with format module, but optional to use.

Maybe it'd be good if you just do simple PR first, adding extension points you would need, and we can go from there?

knoguchi:
I totally understand your thoughts.  We are on the same page.     There is once thing I would like to try before making invasive changes. The .desc file itself is a PB defined by [this proto](https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto).  It means we should be able to load the .desc file out of the box without `protobuf-java`.   Another dependency, protobuf-dynamic, is just a "nice to have" thin wrapper.   No extra jars needed if this idea works.

cowtowncoder:
Excellent! I like the idea even more.

Issue #73, opened 2017-04-04
Cannot resolve inner types in protoc definitions


knoguchi:
When messages are defined as below, the parser failed to resolve t1.i1 type.

```
package mypackage;

message t1 {
        message i1 {
                optional uint32 x = 1;
                optional uint32 y = 2;
        }
}

message t2 {
        optional t1.i1 z = 1;
}
```

Error

```
Exception in thread "main" java.lang.IllegalArgumentException: Unknown protobuf field type 't1.i1' for field 'z' of MessageType 't2' (known enum types: ; known message types: t1, t2)
	at com.fasterxml.jackson.dataformat.protobuf.schema.TypeResolver._resolve(TypeResolver.java:141)
	at com.fasterxml.jackson.dataformat.protobuf.schema.TypeResolver.resolve(TypeResolver.java:93)
	at com.fasterxml.jackson.dataformat.protobuf.schema.NativeProtobufSchema.forType(NativeProtobufSchema.java:67)
	at com.fasterxml.jackson.dataformat.protobuf.schema.ProtobufSchemaLoader.load(ProtobufSchemaLoader.java:51)
```

cowtowncoder:
Unfortunately I can not reproduce this: if I load schema you specify, there is no error.
This could either be due to version you are using (if it's old, could be fixed), or your code does something different that what I have. My test has:

```java
 ProtobufSchema schema = MAPPER.schemaLoader()
                .load(new StringReader(SCHEMA_STR));
```

with schema s above.

knoguchi:
Thanks for the quick response.   I simplified my test code.   The only difference is that I specified the rootType "t2".   I'm using the HEAD of the 2.8 branch pulled today.

```
ProtobufSchema schema = MAPPER.schemaLoader()
                .load(new StringReader(SCHEMA_STR), "t2");
```

And then the stacktrace

```
Exception in thread "main" java.lang.IllegalArgumentException: Unknown protobuf field type 't1.i1' for field 'z' of MessageType 't2' (known enum types: ; known message types: t1, t2)
	at com.fasterxml.jackson.dataformat.protobuf.schema.TypeResolver._resolve(TypeResolver.java:141)
	at com.fasterxml.jackson.dataformat.protobuf.schema.TypeResolver.resolve(TypeResolver.java:94)
	at com.fasterxml.jackson.dataformat.protobuf.schema.NativeProtobufSchema.forType(NativeProtobufSchema.java:67)
[...]

PR #87, opened 2017-05-08
add descriptor loader

knoguchi:
This is a preliminary work for the  #78 "create ProtoSchema from descriptor file".

A "descriptor file" is a set of descriptors.  A descriptor is a binary representation of the .proto file.  Typically one (or more) of the descriptors is the main, and the others are dependencies.

This reader creates a `NativeProtobufSchema` object from the main descriptor.  Dependencies are "copied" into the main descriptor before generating the object.  I chose copying over implementing the import mechanism per #86 discussion. The import statement is a potential security risk, and it would certainly add the complexity.  

This descriptor reader should work for simple message types that do not have references to other nested message types.  #73 needs to be fixed.

Note: descriptor.proto was [copied from Google Protobuf](https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto).   The `reserved` fields were commented out because ProtoParser could not parse.  See #84.


protobuf/src/main/java/com/fasterxml/jackson/dataformat/protobuf/schema/DescriptorReader.java
```diff
@@ -0,0 +1,200 @@
...
+        InputStream in = getClass().getResourceAsStream(DESCRIPTOR_PROTO);
+        descriptorFileSchema = ProtobufSchemaLoader.std.load(in, "FileDescriptorSet");
+        in.close();
+    }
+
+    public Map<String, ProtoFile> readFileDescriptorSet(String descriptorFilePath) throws IOException
+    {
+        InputStream fin;
+        fin = this.getClass().getClassLoader().getResourceAsStream(descriptorFilePath);
+        if (fin == null) {
```
cowtowncoder: If the resource can not be read, why would accessing it as `URL` work differently (and should it)? Maybe at least add a comment explaining what is the motivation here

protobuf/src/test/java/com/fasterxml/jackson/dataformat/protobuf/schema/DescriptorReaderTest.java
```diff
@@ -0,0 +1,26 @@
...
+{
+    public void testParsing() throws Exception
+    {
+        DescriptorReader reader = new DescriptorReader();
+        assertNotNull(reader);
+        Map<String, ProtoFile> protoFileMap = reader.readFileDescriptorSet("main.desc");
+        NativeProtobufSchema nativeSchema = NativeProtobufSchema.construct(protoFileMap.get("main.proto"));
+
+        ProtobufSchema schema = nativeSchema.forType("main1");
+        System.out.println(schema.toString());
```
cowtowncoder: This should be removed.

cowtowncoder:
Ok, I think this could be a good starting point. There are some things I'd probably change, such as making `ProtobufMapper` initiate work instead of reading things in constructor (which is typically not something to avoid), but in general I think it makes sense. This would also remove the dependency from reader to constructing mapper which seems wrong way around.
I think it also would make sense to take different kinds of input sources: assuming a String defines a resource path seems limiting; it may be one option but I assume there are other cases such as reading from an external file.

But one question first: what is intended usage pattern for `ProtoFile`s? I'd prefer not exposing 3rd party types as-is through API, if they can be encapsulated somehow. Can they be converted to protoc definition? Test seems to just verify that something was read, but I assume these are useful for some actual usage.

protobuf/src/main/java/com/fasterxml/jackson/dataformat/protobuf/schema/DescriptorReader.java
knoguchi: I was trying to make it dual purpose -- load the descriptor from resources directory as well as external URL.    I will remove the resources.  I guess the descriptors in the resources directory is only useful for unit tests.

knoguchi:
Thanks for the comments.   The reason why I exposed `ProtoFile`s map is to let the driver to pick the main proto file, then call `NativeProtobufSchema.construct(ProtoFile)` API.  
I will modify it to be something like below to hide the 3rd party API.

```
// in the DescriptorReader class
private Map<String, ProtoFile> protoFileMap;

NativeProtobufSchema forProto(String protoName) {
    return protoFileMap.get(protoName);
}
```

`ProtoFile.toSchema()` method can generate the .proto string but I did not intend the usage.

The test case barely shows that the two .proto files (main.proto and other.proto) are merged.  I will add tests that actually check the correctness of the output.

Other items that you pointed out:
1. I will move the ProtobufMapper for the descriptor schema out of the reader class.
2. add input sources:
    URL
    InputStream
    byte[]

cowtowncoder:
Unfortunately haven't had time to spend on this, but wanted to add a note to say it's still on my radar.

protobuf/src/main/java/com/fasterxml/jackson/dataformat/protobuf/schema/DescriptorLoader.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+            return descriptorMapper.readerFor(FileDescriptorSet.class)
+                                   .with(descriptorFileSchema)
+                                   .readValue(in);
+        }
+        finally {
+            try {
+                in.close();
+            }
+            catch (IOException e) {
```
cowtowncoder: Why catch? Shouldn't it just be propagated as-is?

cowtowncoder:
Actually I think I can make some changes here from API perspective, but retain functionality.
Nothing big, just bit of touch up.

cowtowncoder:
Ok so did refactoring, which is sizable by lines of code. But didn't much change functionality: mostly just changes access point, which is now via `ProtobufMapper`.

PR #102, opened 2017-08-04
[ion] Make IonValueModule public for use outside of IonValueMapper

mcliedtke:
Catching the ion branch up with some changes made internally at Amazon. There are several more changes but I am fairly new to github and this is my first pull request so I wanted to give a small change a trial run.


cowtowncoder:
Makes sense, will be happy to merge.

Was about to ask for CLA, but then realized that this covered by the recent update, thank you for handling this in advcen!.

PR #114, opened 2017-10-16
copyStructure(): avoid duplicate tags when copying tagged binary.

philipa:
`_handleTaggedArray` resets `_tagValue`, so I assume `_handleTaggedBinary()` should too.


cowtowncoder:
Sounds good, thanks! (and bonus points for doing this for 2.9 branch)

PR #115, opened 2017-10-19
Should skip the positive byte which is the last byte of an varint

marsqing:
Should skip the positive byte which is the last byte of an varint


marsqing:
ignore this

PR #125, opened 2017-12-01
Avoid integer overflow when parsing cbor negative numbers

### LINKED ISSUE #124: [cbor] Invalid value returned for negative int32 where the absolute value is > 2^31 - 1
In https://github.com/FasterXML/jackson-dataformats-binary/issues/30, parsing of positive integers was fixed for the case where the uint32 value overflows a java int: https://github.com/FasterXML/jackson-dataformats-binary/blob/d8f82c788c4d8c41c59ca143caf2f989d0e9a9ae/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L679-L689

However, the parsing of negative numbers (major type 1) was not adjusted correctly:

https://github.com/FasterXML/jackson-dataformats-binary/blob/d8f82c788c4d8c41c59ca143caf2f989d0e9a9ae/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L722-L731

The cast of `_numberLong = ((long) v) + -1L;` does not adequately prevent overflow; if v is negative, it has to be adjusted to represent the uint32 value, and then negated:

```
case 2:
    {
        int v = _decode32Bits();
        if (v < 0) {
            long unsignedV = (long) v & 0xFFFFFFFFL;
            _numberLong = -1L - unsignedV;
            _numTypesValid = NR_LONG;
        } else {
            _numberInt = -v - 1;
        }
    }
    break;
```

Similar adjustment is required for other sizes.

As a test case, deserializing `3A 9FF0947F` should produce `-2683344000`; currently it results in `-1611623298`.

JacekLach:
Fixes #124 

I couldnt get the master branch to build:

```
[FATAL] Non-resolvable parent POM for com.fasterxml.jackson.dataformat:jackson-dataformats-binary:3.0.0-SNAPSHOT: Could not find artifact com.fasterxml.jackson:jackson-base:pom:3.0.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 3, column 11
```

So I made this against 2.9. I expect it should be easy to move to master if desired.

### DIFF
--- a/cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/parse/ParserNumbersTest.java
+++ b/cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/parse/ParserNumbersTest.java
@@ -21,7 +21,7 @@ import com.fasterxml.jackson.dataformat.cbor.util.ThrottledInputStream;
 public class ParserNumbersTest extends CBORTestBase
 {
     private final CBORFactory CBOR_F = cborFactory();
-    
+
     public void testIntValues() throws Exception
     {
         // first, single-byte
@@ -101,6 +101,17 @@ public class ParserNumbersTest extends CBORTestBase
         assertEquals(exp, p.getLongValue());
         assertEquals(NumberType.LONG, p.getNumberType());
         p.close();
+
+        // and, combined, a negative number where the mantissa overflows a signed int32
+        input = new byte[] {
+                (byte) CBORConstants.PREFIX_TYPE_INT_NEG + 26, // int32, that is, 4 more bytes
+                -1, -1, -1, -1
+        };
+        p = cborParser(input);
+        assertToken(JsonToken.VALUE_NUMBER_INT, p.nextToken());
+        assertEquals(-1L - 0xFFFFFFFFL, p.getLongValue());
+        assertEquals(NumberType.LONG, p.getNumberType());
+        p.close();
     }
 
     public void testLongValues() throws Exception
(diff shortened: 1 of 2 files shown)


cowtowncoder:
To build master you need to build some of the dependencies locally. I can try to see if there's a way to push snapshot builds of some of the deps.
But PR for 2.9 makes sense regardless since I can merge forward easily (but not back from master).

Thank you for submitting this.

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/parse/ParserNumbersTest.java
```diff
@@ -170,6 +181,21 @@ public void testInt64Overflow() throws Exception
         assertEquals(exp, p.getBigIntegerValue());
         assertEquals(NumberType.BIG_INTEGER, p.getNumberType());
         p.close();
+
+        // and, combined, a negative number where the mantissa overflows a signed int32
+        input = new byte[] {
+                (byte) CBORConstants.PREFIX_TYPE_INT_NEG + 27, // int32, that is, 4 more bytes
```
cowtowncoder: Wouldn't this be `int64`?

cowtowncoder:
Ok, I will have to read this through a few times to fully grok it. I assume tests make sense but it seems like there should be slightly simpler coercions for edge cases (avoid one operation at least).

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/parse/ParserNumbersTest.java
JacekLach: You're right, bad copy paste. I fixed the constant but not the comment. will update later

JacekLach:
Possibly; I just did the same thing that was done for the positive numbers.

PR #142, opened 2018-08-02
TokenBuffer.copyCurrentEvent does not expect getNumberType return null.

mmilkin:
ObjectDeserialization fails with a NPE when a float exists in the ion structure.

There may be other problems when getNumberType() returns null since TokenBuffer.copyCurrentEvent does not expect nulls.

https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/util/TokenBuffer.java#L1054-L1062

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -245,6 +245,8 @@ public class IonParser
                 } else {
                     return NumberType.INT;
                 }
+            case FLOAT:
+                return NumberType.DOUBLE;
             default:
             }
         }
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Thank you for reporting this problem and contributing a fix. Sounds like this is a flaw indeed.
I am on vacation now but will be back in 2 weeks and will look at contribution then.

For merging we will need a filled CLA from https://github.com/FasterXML/jackson filled, signed, scanned and sent to `info` at fasterxml dot com.

mmilkin:
Thanks for the quick response, I will see what I can do about the license.

I also am not sure that this is the correct fix, there is something to be said for having a intelligent default for that switch other then `null`, either a DECIMAL or DOUBLE.

hyandell:
This is an Amazon owned contribution. I've approved/given permission.

mmilkin:
I realize we like precision, changing return type to DOUBLE.

mmilkin:
Any update on this?

cowtowncoder:
@hyandell thanks! @mmilkin Thanks for the ping, I was on vacation and missed approval note.

Issue #143, opened 2018-08-08
[avro] add Stringable class method to SchemaHelper


trompa:
I have a POJO with a Map<Integer, String> that cannot be serialized because Integer is not marked as Stringable.

There is already a @Stringable annotation, but that wont work for Integer wrapper.

There are 2 posible solutions:
* Add Integer to Stringable classes. [AvroSchemaHelper.java](https://github.com/FasterXML/jackson-dataformats-binary/blob/master/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java) holds a list of Stringable classes. Just add it there. As Integer complies with the needed interface to be stringable ( constructor from String, toString method ).
* A solution like the one in avro, to [add new classes to the STRINGABLE_CLASSES Set](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/reflect/ReflectData.java#L104).

baharclerode:
> Add Integer to Stringable classes. AvroSchemaHelper.java holds a list of Stringable classes. Just add it there. As Integer complies with the needed interface to be stringable ( constructor from String, toString method ).

This is specifically the classes considered by the Apache Avro implementation to be stringable. It should stay in sync with the apache implementation for compatibility reasons. Verify how the apache implementation handles Integer map keys, as that is what would need to be followed.

baharclerode:
I just verified - Integer keys are not handled as maps in Avro-- they follow the "Maps with non-stringable keys" pattern, which are encoded as an array of records with key-value fields:
```{
  "type" : "array",
  "items" : {
    "type" : "record",
    "name" : "PairIntegerString",
    "namespace" : "org.apache.avro.reflect",
    "fields" : [ {
      "name" : "key",
      "type" : "int"
    }, {
      "name" : "value",
      "type" : "string"
    } ]
  },
  "java-class" : "java.util.Map"
}```

A) Maps with non-stringable keys are not supported in Jackson currently
B) Even the apache implementation handles this incorrectly, because it chokes on the `"java-class": "java.util.Map"` when trying to deserialize.

TL;DR:
* Treating integer keys as stringable is against the Avro spec
* Due to bugs in deserialization, the Avro reference implementation can't handle Maps with integer keys.

cowtowncoder:
@baharclerode Thank you for additional investigation here.

[...]

PR #144, opened 2018-08-08
Method to add StringableClasses #143

trompa:
Altho I don't really like to modify this static Set.

Any other options?


baharclerode:
This alters the static global instance for all mappers. If this is going to be alterable, it needs a way to be customized per-mapper.

PR #146, opened 2018-08-19
Added including default null value to Avro schema generation

kucera-jan-cz:
Added:
    new Feature enum AVRO_DEFAULT_ENABLED for enabling null inclusion
    RecordVisitor is now include NullNode if AVRO_DEFAULT_ENABLED is enabled and field is not primitive

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
@@ -43,7 +43,12 @@ public class AvroGenerator extends GeneratorBase
          * of this header is not 100% reliably auto-detectable (while header has distinct marker,
          * "raw" Avro content has no limitations and could theoretically have same pre-amble from data).
          */
-        AVRO_FILE_OUTPUT(false)
+        AVRO_FILE_OUTPUT(false),
+
+        /**
+         * Feature that includes to every object type default value "null" when no real default value is defined.
+         */
+        AVRO_DEFAULT_ENABLED(false)
         ;
 
         protected final boolean _defaultState;
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
@@ -205,7 +205,8 @@ public class RecordVisitor
             }
         }
         JsonNode defaultValue = parseJson(prop.getMetadata().getDefaultValue());
-        if(isDefaultsEnabled && defaultValue == null && !prop.getType().isPrimitive()) {
+        if(isDefaultsEnabled && defaultValue == null
+                && writerSchema.getType() == Type.UNION && writerSchema.getIndexNamed(Type.NULL.getName()) != null) {
             defaultValue = NullNode.getInstance();
         }
         writerSchema = reorderUnionToMatchDefaultType(writerSchema, defaultValue);
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
@@ -5,6 +5,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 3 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
```diff
@@ -190,6 +205,9 @@ public void optionalProperty(String name, JsonFormatVisitable handler,
             }
         }
         JsonNode defaultValue = parseJson(prop.getMetadata().getDefaultValue());
+        if(isDefaultsEnabled && defaultValue == null && !prop.getType().isPrimitive()) {
```
baharclerode: Need to check if the writerSchema is a Union and that it contains NULL as a valid sub-schema in the union:

    writerSchema.getType() == Type.UNION && writerSchema.getIndexNamed(Type.NULL.getName()) != null

This can replace the isPrimitive check.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
```diff
@@ -190,6 +205,9 @@ public void optionalProperty(String name, JsonFormatVisitable handler,
             }
         }
         JsonNode defaultValue = parseJson(prop.getMetadata().getDefaultValue());
```
baharclerode: Since you can no longer omit a default when this feature is on to get no default, we should probably add support for explicitly setting no default if `prop.getMetadata().getDefaultValue()` returns the empty string so that there is a way to specify "no default" even when this feature is enabled.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
kucera-jan-cz: Thanks, replaced condition based on proposal.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
baharclerode: Not if you want the field to be _nullable_, just not have a _null default value_. `@JsonProperty(required=true)` will make the field non-nullable.

kucera-jan-cz: I see. Can usage of `@AvroSchema` might be sufficient for such use case?

kucera-jan-cz: @baharclerode - Frankly I am not sure whether there is some action required from my side. Can you please ping me whether I should rewrite proposed change or my previous comment is sufficient?

baharclerode: `@AvroSchema` should be sufficient-- This looks good to me!

kucera-jan-cz: @cowtowncoder - Is there any activity from my side I should do?

cowtowncoder:
@kucera-jan-cz Apologies for slow follow up. Hope to review soon now, finally.
One request: could this be rebased (or new PR created) for 2.10? `master` targets 3.0.0, and that's quite a ways away: so it'd be great to get this in 2.10 instead (from which I'll merge to `master`).

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
```diff
@@ -43,7 +43,12 @@
          * of this header is not 100% reliably auto-detectable (while header has distinct marker,
          * "raw" Avro content has no limitations and could theoretically have same pre-amble from data).
          */
-        AVRO_FILE_OUTPUT(false)
+        AVRO_FILE_OUTPUT(false),
+
+        /**
+         * Feature that includes to every object type default value "null" when no real default value is defined.
```
cowtowncoder: This should explain that setting only affects Schema generation, and NOT on reading or writing.

cowtowncoder:
Hmmh. Ok, I think I'll merge this in and change naming.

Looks like there's one snag wrt 2.10, so will first just merge to `master`.

PR #156, opened 2019-01-03
fix #155: Only flush outputs when FLUSH_PASSED_TO_STREAM is allowed

### LINKED ISSUE #155: Inconsistent support for FLUSH_PASSED_TO_STREAM
Cbor and Smile serializers attempt to flush streams on close when AUTO_CLOSE_TARGET is disabled, even if FLUSH_PASSED_TO_STREAM is also disabled.

carterkozak:
Previously most serializers would attempt a flush when
AUTO_CLOSE_TARGET was disabled, which is not expected if
FLUSH_PASSED_TO_STREAM is also disabled.

This change results in binary serializers behaving the same as
the json serializer.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/AvroGenerator.java
@@ -292,7 +292,9 @@ public class AvroGenerator extends GeneratorBase
 
     @Override
     public final void flush() throws IOException {
-        _output.flush();
+        if (isEnabled(StreamWriteFeature.FLUSH_PASSED_TO_STREAM)) {
+            _output.flush();
+        }
     }
     
     @Override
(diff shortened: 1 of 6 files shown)


cowtowncoder:
Sounds legit, thank you for submitting this PR!
I wonder if there'd be an easy way to test these via unit tests, to guard against regression?

Also: I think I'll need to backport this in `2.10` (since `master` is now for `3.0.0-SNAPSHOT`), but changes are small enough that should be easy.

cowtowncoder:
Looks like `jackson-core` has trivial test for the feature under `src/test/java/com/fasterxml/jackson/core/main/TestGeneratorClosing.java`, so could cut'n paste.
Unfortunately there's no way (at this point) to easily share test classes.

cakofony:
It should be relatively straightforward to write a test -- I can take a look later tonight. I'd really appreciate a 2.x backport, happy to help out with that.

cowtowncoder:
@cakofony excellent: thank you again for reporting this & providing patch!

cakofony:
Thanks for the quick replies! I've added tests for cbor and smile, but I'm less familiar with the other formats and the same style of template didn't seem to work.

cowtowncoder:
Thank you! I ended up manually merging bits and pieces as I realized that backporting will be challenging otherwise, but change is the same.

PR #171, opened 2019-07-19
JPMS Avro - Static

GedMarc:
to allow dependency to name properly, and allow implementor to define and change as need be

### DIFF
--- a/avro/src/moditect/module-info.java
+++ b/avro/src/moditect/module-info.java
@@ -1,13 +1,13 @@
 // Generated 15-Mar-2019 using Moditect maven plugin
 module com.fasterxml.jackson.dataformat.avro {
-    requires com.fasterxml.jackson.annotation;
+    requires transitive com.fasterxml.jackson.annotation;
     requires com.fasterxml.jackson.core;
     requires com.fasterxml.jackson.databind;
 
     // silly avro Apache impl, its deps:
-    requires avro;
-    requires jackson.core.asl;
-    requires jackson.mapper.asl;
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Ok so for Avro it makes sense (as use of Apache Avro is optional). But is the same true for Ion? I don't think there is currently any way to avoid loading Ion codec.

Also wrt Avro -- I hope to soon upgrade Apache Avro dep to 1.9 which will change things a bit here, as it drops Jackson 1.x (relies now on 2.x). But until then change makes sense of course.

GedMarc:
Ion will be looked for on the UNNAMED module section, and lets consumers specify the module for requirement :) 
As long as the dependency is included (any version) the packages will pick up, 

I did something very similar here (https://github.com/GedMarc/guice-servlet/blob/master/moditect/module-info.java) on the guice module, which has classes for both servlet 2.5 and 3 - 

this allows the consumer to load either of them and not crash on building jlink or something like that

cowtowncoder:
Ok so looks like Avro 1.9 upgrade is not as easy as I hoped (things compile fine, but fail lots of unit tests -- and unit test contributions use great-jazzy test matrix features which means I can not figure out what even fails... which is not that great, but I digress), so it's probably best to prepare for 1.8.x being the implementation for Jackson 2.10.

So I'll merge this: not much Ion usage, and I suspect their module info ... does not exist? In which case someone probably has to repackage it, so `static` makes sense to allow users to work around that gap.

PR #175, opened 2019-08-06
Fix union serialization

### LINKED ISSUE #173: Improve Union type serialization performance
The serialization mechanism currently calls the `ReflectData.get().resolveUnion()` Method for resolving union schemas. This call represents a huge penalty as it generates the entire schema for the object behind the scenes to get the schema name and find the corresponding index. It's even worse because sometimes it not even possible to generate the schema for the actual value, causing the serialization to fail:

```java
class Foo {
     // The schema for this object cannot be generated, as the type of tricks can be anything.
     // However, for the sake of the serialization, it doesn't matter, as the type has been specified  in the schema
     public Set<?> tricks = new HashSet<>();
}
```
This issue can be resolved by checking the set of union schema using the class name instead of using `ReflectData.get().resolveUnion()`. There is [already a snippet in the code that does exactly this,](https://github.com/FasterXML/jackson-dataformats-binary/blob/2f29c000b67f2dbe0c234b05d0fb65f43f5b1a55/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/ser/AvroWriteContext.java#L291) but is currently disabled.

marcospassos:
Fixes #173

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/ser/AvroWriteContext.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/ser/AvroWriteContext.java
@@ -276,8 +276,7 @@ public abstract class AvroWriteContext
             // !!! TODO:
             //  - ByteBuffer
             //  - Number wrappers (Integer, ...)
-            
-            /*
+
             String typeId = AvroSchemaHelper.getTypeId(datum.getClass());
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Looks legit, so with CLA will be able to merge this in to be included in 2.10.0.pr2!

PR #177, opened 2019-08-21
Deserialization of "empty" Records as root values fails

marcospassos:
@cowtowncoder Could you please give me some guidance on how to fix this issue? I'd like to fix it in time for 2.10.

The unit test spots the issue: empty records fail on deserialization.

### DIFF
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/records/RecordWithNoFieldsTest.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/records/RecordWithNoFieldsTest.java
@@ -1,0 +1,42 @@
+package com.fasterxml.jackson.dataformat.avro.interop.records;
+
+import java.io.IOException;
+import org.apache.avro.Schema;
+import org.apache.avro.SchemaBuilder;
+import org.codehaus.jackson.annotate.JsonAutoDetect;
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Quick question here: is this related to how Apache Avro library generates schema, or can this be easily replicated with textual schema and/or Avro module's schema generator?

Just figuring out where to add the test: if it's related to apache lib, interop is fine; but if not I try to keep depenencies to it to minimum.

I will try to help with this issue over the weekend: thank you for all your work on Avro module & apologies for limited time I have to spend to help keep you unblocked.

marcospassos:
> Quick question here: is this related to how Apache Avro library generates the schema, or can this be easily replicated with textual schema and/or Avro module's schema generator?

I don't believe that is related to schema generation, but rather to the serialization logic that expects a non-empty list of field, mainly because I’m providing the schema in the test.

Thank you for your support. I'm glad to contribute to a library that helps us that much.

cowtowncoder:
@marcospassos Ok so the test can be simplified to use (f.ex) embedded schema it sounds like.

cowtowncoder:
Did not have time to dig into this today: it's at top of my list and (aside from adding comments which I prioritize high to try to help everyone), and should be able to at least get test added, have a look at exception tomorrow,

PR #178, opened 2019-09-09
Fix issue wit input offsets when parsing CBOR from `InputStream`

iziamos:
This fixes an issue I discovered where CBORParser#_loadToHaveAtLeast didn't increment _currInputProcessed pointer if array move didn't occur.  Observed because calling CBORParser#getCurrentLocation after CBORParser#finishToken returned offsets smaller than expected if a token spanned between two chunks read from the stream.


cowtowncoder:
@iziamos Thank you for reporting the problem, submitting the fix!
I merged this manually so it can go in 2.10 (`master` is for 3.x), and will be in 2.10.0.

PR #182, opened 2019-09-26
[Help needed] Fix custom null serialization

marcospassos:
@cowtowncoder implementing a null custom serialization for an internal library, I stumbled upon a bug that prevents serialized null from being deserialized (which should affect other packages, such as JSR-353). For now, this PR only includes a test that reproduces the problem.

Error message:
```
com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
```

Could you give me some guidance on how to fix it?

### DIFF
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/CustomSerializationTest.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/CustomSerializationTest.java
@@ -1,5 +1,6 @@
 package com.fasterxml.jackson.dataformat.avro.interop.annotations;
 
+import com.fasterxml.jackson.databind.JsonMappingException;
 import java.io.IOException;
 import java.util.Objects;
 import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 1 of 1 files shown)


avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/CustomSerializationTest.java
```diff
@@ -105,4 +106,61 @@ public void testUnionCustomDeSerialization() throws IOException {
...
+            }
+        }
+
+        public static class Deserializer extends JsonDeserializer<NullValue> {
+            @Override
+            public NullValue deserialize(
+                final JsonParser parser,
+                final DeserializationContext context
+            ) throws IOException {
+                parser.nextToken();
```
cowtowncoder: I think this is not right: should instead check that current token is `JsonToken.VALUE_NULL`: parser already points to the first token of value to deserialize.

cowtowncoder:
It'd be good to maybe file issue separately, link to/from this PR.
There's one small problem with custom deserializer, not sure if it matters.

marcospassos:
> It'd be good to maybe file issue separately, link to/from this PR.
I'll open a PR.

I just pushed a new commit, but it doesn't seem to fix the issue.

marcospassos:
* the same does not happen when serializing/deserializing to/from JSON.

cowtowncoder:
Ok. Thank you for verifying.

I think this could be related to one oddity wrt Avro handling of `null`s: unless null value is part of union, it "does not exist": that is, it does not take any space. Value of a field with only `null type` is always `null`, and nothing is encoded.
This is fine in theory, but presents unique problem at root-level reader: it is not easy to know if stream has ended or not -- just because there are no bytes left does not necessarily mean we are done as `null` may still be implicitly available!
(I had to fix one bug earlier related to this).

marcospassos:
Dammit, now I see why I couldn't find an easy way to fix it myself without consuming your limited time. But it seems a critical bug to fix, as it can break things seriously. Again, let me know if I can help you fix it.

cowtowncoder:
@marcospassos Unfortunately I do not really have time to work on this now.

cowtowncoder:
I don't think this is active any longer; closing (may be re-filed of course)

PR #192, opened 2019-12-18
Allow IonObjectMapper with class name annotation introspector to dese…

ankel:
…rialize non-annotated payloads.

Currently, an Ion Mapper with Classname Id resolver (and likely other kind of type id resolver) loses the ability to deserialize plain, non-annotated payload using only generic hints provided through type reference. This bug was likely because of this commit [1].

The fix here is to construct the type with full generic information instead of just using defaultImpl (which is now just a class).

---
[1] https://github.com/FasterXML/jackson-databind/commit/596c6dd1850fbb92ee34ff64c7bc6e2aafc45207#diff-48b058c7384ce3f9fc9e7440fc7dc804L89


cowtowncoder:
Fix looks legit, will merge.

Issue #167, opened 2019-06-26
Incompatibility with Avro >=1.9.0 (upgrade to Avro 1.11.3)


Sage-Pierce:
Avro version 1.9.0 was released last month (https://mvnrepository.com/artifact/org.apache.avro/avro) and my team has run in to compatibility issues between `jackson-dataformat-avro` and this latest release.

In particular, it looks to be the case that Avro has switched from Codehaus Jackson to FasterXML Jackson, which results in `NoSuchMethodError` thrown when accessing utility methods.

The following test exposes the issue:

```
    @Test
    public void beanShouldBeSerializableAsAvroBinaryAndDeserializable() throws Exception {
        TestBean testBean = new TestBean();
        testBean.setData("Data");

        AvroMapper avroMapper = new AvroMapper();

        AvroSchema avroSchema = avroMapper.schemaFor(TestBean.class);

        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
        avroMapper.writer().with(avroSchema).writeValue(outputStream, testBean);

        byte[] serialized = outputStream.toByteArray();

        assertNotNull(serialized);
        assertTrue(serialized.length > 0);

        TestBean deserialized = avroMapper.readerFor(TestBean.class).with(avroSchema).readValue(serialized);

        assertEquals(testBean, deserialized);
    }

    public static final class TestBean {

        private String data;

        @Override
        public boolean equals(Object o) {
            if (this == o)
                return true;
            if (o == null || getClass() != o.getClass())
                return false;
            TestBean testBean = (TestBean) o;
            return Objects.equals(data, testBean.data);
        }

        @Override
        public int hashCode() {
            return Objects.hash(data);
        }

        public String getData() {
            return data;
        }

        public void setData(String data) {
            this.data = data;
        }
    }
```

This test passes on v1.8.2 of Avro, but fails with v1.9.0 on the following:
```
java.lang.NoSuchMethodError: org.apache.avro.util.internal.JacksonUtils.toObject(Lorg/codehaus/jackson/JsonNode;)Ljava/lang/Object;

	at com.fasterxml.jackson.dataformat.avro.schema.RecordVisitor.schemaFieldForWriter(RecordVisitor.java:192)
	at com.fasterxml.jackson.dataformat.avro.schema.RecordVisitor.optionalProperty(RecordVisitor.java:121)
[...]

Issue #145, opened 2018-08-10
Add `AvroGenerator.Feature.ADD_NULL_AS_DEFAULT_VALUE_IN_SCHEMA` for adding default `null` in generated schema


kucera-jan-cz:
Hello,
currently we are using for generation schema from POJO ReflectData.AllowNull.get() however we would like to switch to jackson-dataformats-binary. The only issue with the switch is absence of "default" value in schema (which we currently have in our schema registry and would break our compatibility). Is it possible to set this behavior through AvroFactory or AvroSchemaGenerator?

For clarification here is example from jackson-dataformats-binary:
`{"type":"record","name":"PersonV1","namespace":"dummy","fields":[{"name":"age","type":["null","string"]}]}`
And here example from ReflectData.AllowNull.get():
`{"type":"record","name":"PersonV1","namespace":"dummy","fields":[{"name":"age","type":["null","string"],"default":null}]}`

Thanks in advance for your answer

cowtowncoder:
@kucera-jan-cz Could you include bit more information on code you are using: I assume you have a simple POJO definition (class `PersonV1`), and then create `AvroSchema` from it. But how are you serializing it to get output?

kucera-jan-cz:
@cowtowncoder Hello and thanks for your reply.
Yes, I am having simple POJO and serializing it with our own serializer with following code:
```
           int registeredSchemaId = this.schemaRegistry.register(subject, schema);

            ByteArrayOutputStream out = new ByteArrayOutputStream();
            out.write(0);
            out.write(ByteBuffer.allocate(4).putInt(registeredSchemaId).array());

            DatumWriter<Object> dw = new ReflectDatumWriter<>(schema);
            Encoder encoder = ENCODER_FACTORY.directBinaryEncoder(out, null);
            dw.write(value, encoder);
            encoder.flush();
            return out.toByteArray();
```

baharclerode:
Have you tried using `@AvroDefault("null")` or `JsonProperty(defaultValue = "null")` ?

cowtowncoder:
@kucera-jan-cz I meant Jackson code used to produce output that is missing "default" entry. I am not sure I fully understand exact steps of desired Jackson dataformat solution.

baharclerode:
Oh, I see what you're asking. Currently, Jackson will generate unions for optional fields:

```java
package dummy;

public class PersonV1 {

    public PersonV1(String age) {
        this.age = age;
    }

    private String age;

[...]

PR #195, opened 2020-02-21
[Avro] Remove dependencies upon Jackson 1.X and Avro's JacksonUtils

baharclerode:
This is the first half of a fix for #167 

This removes the references to Jackson 1.x and `JacksonUtils` (which is internal to Avro), allowing Jackson-Avro to be binary-compatible Avro 1.9 or Avro 1.8 but not semantically-compatible in all cases, particularly around nested classes; For simple use-cases, this is sufficient to allow Jackson-Avro to work with Avro 1.9, but I would not by any means call it "supported" due to the nasty gotchas that still lurk.

This does *not* provide full backwards and forwards compatibility with Avro 1.9; Notably:
* Generating a schema using Jackson-Avro w/ Avro 1.9 will still produce a 1.8 schema for nested classes; Avro 1.9 can't use these for serialization (But Jackson-Avro w/ Avro 1.9 can). Deserialization still seems to work.
* Uses of the `@AvroAlias` annotation to create aliases to nested types will break, since previously these had to follow the Avro 1.8 conventions for namespacing nested types
* **Mixing of Avro 1.8 and Avro 1.9 anywhere is right out.** You can't generate a schema with Avro 1.8 and have Jackson-Avro w/ Avro 1.9 do anything with it if nested classes are involved, and vice versa. The same goes for using Jackson-Avro w/ Avro 1.8 to generate schemas and Avro 1.9 to serialize/deserialize, and vice versa. Don't try it. If it works, it'll work just long enough for you to forget about it not supporting nested classes.

Fixing of the above will probably require porting all of the Avro union resolver code into Jackson-Avro with modifications to allow it to resolve both 1.8- and 1.9-styled namespaces, and a mapper flag that indicates if you want 1.8 or 1.9-compatibility for `@AvroAlias` annotations and generated schemas.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
@@ -14,8 +14,8 @@ import org.apache.avro.reflect.Stringable;
 import org.apache.avro.specific.SpecificData;
 
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.databind.BeanDescription;
-import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.*;
 import com.fasterxml.jackson.databind.introspect.AnnotatedClass;
 import com.fasterxml.jackson.databind.introspect.AnnotatedConstructor;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonFormatTypes;
@@ -23,6 +23,11 @@ import com.fasterxml.jackson.databind.util.ClassUtil;
 
 public abstract class AvroSchemaHelper
 {
+    /**
+     * Dedicated mapper for handling default values (String &lt;-&gt; JsonNode &lt;-&gt; Object)
+     */
+    private static final ObjectMapper DEFAULT_VALUE_MAPPER = new ObjectMapper();
+
     /**
      * Constant used by native Avro Schemas for indicating more specific
      * physical class of a value; referenced indirectly to reduce direct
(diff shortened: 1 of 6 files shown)


cowtowncoder:
Thanks! I'll make a minor post-merge change probably to isolate use of `ObjectMapper`, but other than that it's the first big step.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
```diff
@@ -332,9 +341,44 @@ public static String getFullName(Schema schema) {
                 return namespace + name;
             }
             StringBuilder sb = new StringBuilder(1 + namespace.length() + name.length());
-            return sb.append(namespace).append('.').append(name).toString();
-        default:
+            // Check if this is a nested class
+            String nestedClassName = sb.append(namespace).append('$').append(name).toString();
+            try {
+                Class.forName(nestedClassName);
```
cowtowncoder: @baharclerode assuming most common case is NOT inner class, would it make sense to try loading "regular" class first, inverting logic? Only asking since there's quite a bit of overhead for exception handling and possibly other corner cases for trying to load non-existing class.

cowtowncoder:
Merged; did some minor renaming.

cowtowncoder:
Turns out I had to do some more tweaking in 3.0, to avoid breaking #145. Not sure if similar "null masking" might be needed for 2.x, come to think of that (that is, when converting from `NullNode`, should use `JsonProperties.NULL_VALUE`, to retain "explicit null as default" vs "no default value).

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
baharclerode: I don't think so; Either way, we're always doing at least 1 `Class.forName()`; Either we check for the non-nested class, and assume if that fails that it's nested, or we always check for the nested class, and if that fails, assume that it's not-nested.

cowtowncoder: Alas, looks like this lead to a major performance issue for some usage (see #219 for attempts to work around the issue). I assume that case of finding class is relatively cheap, but that failure is enormously expensive (as that is the "rare" case being optimized against), depending on complexity of classpath for application.

PR #200, opened 2020-03-28
[cbor] Write large numbers in uint32 form where possible

yawkat:
When minimal ints are enabled, the generator will now encode numbers that exceed java int range but still can be encoded as 5 bytes in cbor as such.


cowtowncoder:
@yawkat I am not 100% sure I understand the issue being fixed here -- would it be possible to create issue to explain the problem (and to include in release notes)?

yawkat:
I've opened #201. That issue also includes some other ideas on efficient integer representation.

cowtowncoder:
I ended up merging the solution manually; same checks, test (but added a few other tests for complteness); closing this. Thank you for submitting it!

PR #204, opened 2020-05-06
Add getIonSystem() accessor in IonFactory

paulfferraro:
It is sometimes necessary to access the underlying IonSystem when interacting
with an IonObjectMapper or ObjectWriter (eg. to replicate the logic in
IonObjectMapper#writeValueAsIonValue()).

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonFactory.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonFactory.java
@@ -135,6 +135,10 @@ public class IonFactory extends JsonFactory {
         return new IonFactory(this, null);
     }
 
+    public IonSystem getIonSystem() {
+        return _system;
+    }
+
     @Override
     public Version version() {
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Makes sense, I can add this.
Just one request: could you rebase against 2.11? In theory, being API add, should go in next minor version (2.12), but I think this is small enough thing it can go in 2.11.1. And since 2.10 is going in maintenance mode, I think `2.11` branch is the best place (I'll document it as addition in official 2.12, but versions 2.11.1 and up do contain it).

paulfferraro:
Any chance I can convince you to push to 2.10 and merge to 2.11? Or back-port to 2.10? Many parts of the org just recently updated to jackson-databind-2.10 and I want to be able to bring the Ion support up to speed instead of using the old 2.8 version of jackson-over-ion support we currently have.

cowtowncoder:
I guess that is ok (just wrt "unofficial" add -- will not be removed, just not documented), I don't see harm in there.

PR #208, opened 2020-06-14
Add support to (de)serialize Ion timestamps with java.time classes

mcliedtke:
Adding support for some java 8 `java.time` classes to (de)serialize to [Ion's timestamp type](http://amzn.github.io/ion-docs/docs/spec.html#timestamp).

Starting with just the following since it seemed pretty clear cut that there could be support:
* `java.time.Instant`
* `java.time.OffsetDateTime`
* `java.time.ZonedDateTime`

I was less sure about how to handle `java.time.LocalDate` and `java.time.LocalDateTime` so they have been omitted. 

Serialization is fairly straightforward, each class is pretty easily transformed into an Ion timestamp. The caveat is that the transformation for a `java.time.ZonedDateTime` will be slightly lossy since an Ion timestamp cannot represent the timezone, just the offset. In this case the local offset is used.

Deserialization is also mostly straight forward. In the event that the Ion timestamp's time local offset [is unknown (-00:00)](https://www.ietf.org/rfc/rfc2822.txt), then the default timezone is used from the context. 

(De)serialization to/from an Ion int or Ion decimal is added as well though I haven't tried to add support to/from an Ion string.

---

Some of the code could be refined with lambdas and function references in the future if this module is moved to java 8.

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/jsr310/IonJavaTimeModule.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/jsr310/IonJavaTimeModule.java
@@ -1,0 +1,40 @@
+package com.fasterxml.jackson.dataformat.ion.jsr310;
+
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.ZonedDateTime;
+
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.core.json.PackageVersion;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+
+public class IonJavaTimeModule extends SimpleModule {
+
+    private static final long serialVersionUID = 1L;
+
+    public IonJavaTimeModule() {
+        super(PackageVersion.VERSION);
+        addSerializer(Instant.class, IonTimestampInstantSerializer.INSTANT);
+        addSerializer(OffsetDateTime.class, IonTimestampInstantSerializer.OFFSET_DATE_TIME);
+        addSerializer(ZonedDateTime.class, IonTimestampInstantSerializer.ZONED_DATE_TIME);
+
+        addDeserializer(Instant.class, IonTimestampInstantDeserializer.INSTANT);
+        addDeserializer(OffsetDateTime.class, IonTimestampInstantDeserializer.OFFSET_DATE_TIME);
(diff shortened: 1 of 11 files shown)


cowtowncoder:
Sounds reasonable; I assume you know what you are doing :)
Since I think we have CCLA for Amazon, including you that should be fine.

Couple of questions:

1. Ion module is still nominally JDK7 compatible before this change, I think, so to merge it in, module should be considered JDK8-only. I think this should mean that fix goes in 2.12.0, as 2.11.0 was already released? If so, I think it'd make sense to rebase this against 2.12 branch
2. Since you have contributed changes and know your way around, would you like to become the maintainer of the Ion format? I could give you bit more access, add information on README (and other places where we note owners of various modules), and would know to include you for code reviews for PRs others submit

mcliedtke:
1. Yeah I think the module is currently JDK7 compatible. Would a JDK7 consumer be able to consume these changes safely by just avoiding the IonJavaTimeModule and other classes? But in general I agree with your point, it would be better to just make this entire module require JDK8 and that would be better to target for 2.12. What's the timeline look like for 2.12? 2.11 was just released so I presume maybe later this year?
2. Yes, I can make some updates to the README.md
3. I would be excited to get more involved and take on some responsibilities. I am up for being *a* maintainer but not so sure about being *the* maintainer :). I am passionate about Ion and Jackson but won't claim to know Jackson thoroughly. Would I still be able to delegate/defer to you in certain areas? I probably should also check internally to see if there is anything I need to be aware of.

cowtowncoder:
So, on 2.12: I am hoping to make this release "quick one" similar to 2.11, but it is probably still couple of months out. If I had to guess, first RC by end of August, release by end of Sep or early Oct.

On being maintainer: I would still handle release process, help merge changes from databind and so on so I would not just dump maintenance on you. It'd be more that I would rely on you to validate Ion-related parts; I could help from overall Jackson design aspects.
This is much easier here than with, say, Scala and Kotlin modules where my expertise is much more limited.

mcliedtke:
Opened a new PR with the changes against the 2.12 branch: https://github.com/FasterXML/jackson-dataformats-binary/pull/213

I am working out what permsions I need for being a maintainer and will reach out once I hear more, this is still something I would be very interested in.

PR #210, opened 2020-06-21
Fix bug when deserializing Avro data using schema evolution 

marcospassos:
This little typo caused us a serious problem when deploying to production a new version last week. We took several hours until we managed to track this bug down. It was hard to debug because it involves schema evolution.

When combining two schemas, if one of them has a map field, you got the following error:

```
Not an array: {“type”:“map”,“values”:..}
```

@cowtowncoder any chances of releases a patch version anytime soon? It'd help us a lot.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroReaderFactory.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroReaderFactory.java
@@ -288,7 +288,7 @@ public abstract class AvroReaderFactory
                 String valueTypeId = readerSchema.getValueType().getProp(AvroSchemaHelper.AVRO_SCHEMA_PROP_CLASS);
                 return MapReader.construct(dec, typeId, keyTypeId, valueTypeId);
             }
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Would it be possible to add a unit test to verify and guard against the issue?

Also, this seems to be against `master`, but that is for 3.0.0-SNAPSHOT, so it probably needs to be against some other branch -- 2.10 assuming code hasn't changed (and I can merge forward); I am hoping to release 2.11.1 quite soon (2.10.5 probably will take time as there aren't that many fixes accumulated).

marcospassos:
Sure, I will open a PR against 2.10 with unit tests. Any chances of
releasing a patch version next week?

marcospassos:
Surpassed by #211

cowtowncoder:
Ok thanks, I will try to get this merged soon.

As to release, we'll see, I might get to releasing 2.11.1 over next weekend if all goes well.
Maintaining everything with Jackson is so much work that it is getting hard to find time and inspiration to do this solely as community service, which is what it comes down to at this point. :-(

PR #211, opened 2020-06-22
Fix schema evolution involving maps of non-scalar

marcospassos:
Rework on branch 2.10, surpasses #210.

@cowtowncoder I've added tests to prevent regression as you asked.

Just reiterating, if possible, releasing a patch anytime soon would help us a lot :)

 Please, let me know if you need anything else.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroReaderFactory.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/AvroReaderFactory.java
@@ -289,7 +289,7 @@ public abstract class AvroReaderFactory
                 String valueTypeId = readerSchema.getValueType().getProp(AvroSchemaHelper.AVRO_SCHEMA_PROP_CLASS);
                 return MapReader.construct(dec, typeId, keyTypeId, valueTypeId);
             }
-            return MapReader.construct(createReader(writerElementType, readerSchema.getElementType()), typeId, keyTypeId);
+            return MapReader.construct(createReader(writerElementType, readerSchema.getValueType()), typeId, keyTypeId);
         }
 
         protected AvroStructureReader createRecordReader(Schema writerSchema, Schema readerSchema)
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Merged. The reason I am not sure about 2.10 release is just low number of fixes as per

https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.10.5

(and since this is likely the last full patch set for 2.10)

whereas many more for 2.11:

https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.11.1

and so if possible it'd be great to verify that 2.10->2.11 could work for your use case.
Every new full patch release takes 2-3 hours for me so I will need to balance time available on cutting releases.

cowtowncoder:
Hmmmh. Looks like this change broke 2.12 tests :-(

cowtowncoder:
.... or maybe it was already broken. Need to check into failure. Will add a note if I find something related to this change.

marcospassos:
both 2.10 and 2.12 works for us, we just need a released version including the fix, so choose whatever requires less time of yours :)

About the broken tests, it's weird because the fix is a one-line change that is clearly a mistake (the test is the proof).

cowtowncoder:
@marcospassos I think it is unrelated to this change, apologies for false alarm. It is related to coercion config changes in 2.12 somehow, need to figure out exactly how.

Thanks wrt 2.10/2.11; just wanted to know if you had specific constraints at this point.

cowtowncoder:
Ah. Interesting. Databind wrongly prevent coercion from "Blank string" for `Character`, in case of a single white space characted. Something databind did not test but Avro module does.... Nice indirect catch. :)

marcospassos:
Hey @cowtowncoder! Any idea on when it will be released?

cowtowncoder:
@marcospassos I'll have to think of when to release 2.10.5: it might be the last 2.10 patch release and there haven't been that many fixes. So there's some balancing wrt how long to wait.

But I think this fix was included in 2.11.1 which has been released, if you can use that version?

PR #213, opened 2020-06-27
Add support to (de)serialize Ion timestamps with java.time classes

mcliedtke:
(continued from: https://github.com/FasterXML/jackson-dataformats-binary/pull/208 - since I switched branches I opened a different PR)

Adding support for some java 8 `java.time` classes to (de)serialize to [Ion's timestamp type](http://amzn.github.io/ion-docs/docs/spec.html#timestamp).

Starting with just the following since it seemed pretty clear cut that there could be support:
* `java.time.Instant`
* `java.time.OffsetDateTime`
* `java.time.ZonedDateTime`

I was less sure about how to handle `java.time.LocalDate` and `java.time.LocalDateTime` so they have been omitted. 

Serialization is fairly straightforward, each class is pretty easily transformed into an Ion timestamp. The caveat is that the transformation for a `java.time.ZonedDateTime` will be slightly lossy since an Ion timestamp cannot represent the timezone, just the offset. In this case the local offset is used.

Deserialization is also mostly straight forward. In the event that the Ion timestamp's time local offset [is unknown (-00:00)](https://www.ietf.org/rfc/rfc2822.txt), then the default timezone is used from the context. 

(De)serialization to/from an Ion int or Ion decimal is added as well though I haven't tried to add support to/from an Ion string.

---

Some of the code could be refined with lambdas and function references in the future if this module is moved to java 8.

### DIFF
--- a/ion/README.md
+++ b/ion/README.md
@@ -32,6 +32,17 @@ byte[] encoded = mapper.writeValueAsBytes(value);
 SomeType otherValue = mapper.readValue(data, SomeType.class);
 ```
 
+### java.time JSR 310
+There is support for (de)serializing some `java.time` classes directly from/to Ion timestamp values.
+
+```java
+IonObjectMapper mapper = IonObjectMapper.builder()
+        .addModule(new IonJavaTimeModule())
+        //Disable writing dates as numeric timestamp values to allow writing as Ion timestamp values. 
+        .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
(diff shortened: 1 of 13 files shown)


cowtowncoder:
Merged for 2.12, will add info on release notes

cowtowncoder:
Updated README, and also created `dataformat-ion` team (with @mcliedtke invited) to give push access (it'll be for whole repo, technically, although conceptually for Ion related changes).

PR #220, opened 2020-09-14
implement property and optionalProperty(String, JsonFormatVisitable, JavaType) methods 

jwijgerd:
implemented the propery methods that take property name and type as inputs. This is needed to support custom (de)serializers that are using these methods. This can only be used with the DefaultTagGenerator so an IllegalStateException will be thrown if this gets mixed with properties annotated with @JsonProperty.index

I created this one on the 2.12 branch because my master branch was not compiling. can be cherry picked to master as well

### DIFF
--- a/protobuf/src/main/java/com/fasterxml/jackson/dataformat/protobuf/schemagen/AnnotationBasedTagGenerator.java
+++ b/protobuf/src/main/java/com/fasterxml/jackson/dataformat/protobuf/schemagen/AnnotationBasedTagGenerator.java
@@ -13,4 +13,9 @@ public class AnnotationBasedTagGenerator implements TagGenerator
         throw new IllegalStateException("No index metadata found for " + writer.getFullName()
                 + " (usually annotated with @JsonProperty.index): either annotate all properties of type " + writer.getWrapperName().getSimpleName() + " with indexes or none at all");
     }
+
+    @Override
+    public int nextTag() {
+        throw new IllegalStateException("Index metadata not found, (usually annotated with @JsonProperty.index): either annotate all properties with indexes or none at all");
+    }
 }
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Thank you for contributing this! I'll have to take a look to make sure I understand it, but sounds reasonable. And 2.12 branch is actually very good choice; I can merge it up to `master` easily and 2.12.0 should be released within maybe a month.

jwijgerd:
Thanks for picking this up so soon. I recently contributed something to jackson-datatype-joda but can't remember signing a CLA so you may want to check that.. Happy to sign if needed!

cowtowncoder:
@jwijgerd I did check our repo, and there's CLA for jackson 2.5, seems like afterburner contribution (which would also mean no need for another for Joda). :)

cowtowncoder:
Quick question: any chance to provide a simple unit test with this? (ideally something failing without this, passing with fix)

The other thing was possible concern wrt adding a new method in interface `TagGenerator`, regarding backwards compatibility, but I guess that can't be helped (until baseline is increased to Java 8 allowing default impl).

jwijgerd:
Sure thing, I will provide a test. Let me think about TagGenerator interface a bit and get back to you

cowtowncoder:
Sounds good! I'll make sure this can get in 2.12.0-rc1 (planning to release over next weekend, in 7 days, if all goes well).

jwijgerd:
@cowtowncoder  Sorry I wasn't able to work on this last week, hope to get the test in today or tomorrow

cowtowncoder:
@jwijgerd np, thank you for update! I am hoping to release 2.12.0-rc1 next weekend, if possible, would be great to include this -- but it is not the hard deadline, can include after, before 2.12.0 final.

cowtowncoder:
@jwijgerd  At this point 2.12.0-rc2 is not yet out, but will be within a week -- so would be great to see if this could still get in?

jwijgerd:
@cowtowncoder will try to do my best, very hard to find some time atm!

cowtowncoder:
@jwijgerd Understood, we are all volunteers, so no pressure.

cowtowncoder:
Just noting that this could be targeted for 2.13 (no hurry)

cowtowncoder:
I don't think this is active any more, will close.

PR #222, opened 2020-09-29
Add `CBORGenerator.Feature.LENIENT_UTF_ENCODING` for lenient handling of Unicode surrogate pairs on writing

guillaumebort:
If enabled, the generator will output the Unicode Replacement Character for invalid unicode sequence (invalid surrogate chars in the Java String) instead of failing with an IllegalArgumentException.

Also this PR remove the code duplication between `_shortUTF8Encode2` and `_encode2`.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
@@ -74,6 +74,8 @@ public class CBORGenerator extends GeneratorBase
          * Feature that determines if an invalid surrogate encoding found in the
          * incoming String should fail with an exception or silently be outputed
          * as the Unicode 'REPLACEMENT CHARACTER' (U+FFFD)
+         *
+         * @since 2.12
          */
         LENIENT_UTF_ENCODING(false),
 
@@ -152,6 +154,11 @@ public class CBORGenerator extends GeneratorBase
 
     protected boolean _cfgMinimalInts;
 
+
+    /**
+     * If true we will output the REPLACEMENT_CHAR for invalid unicode sequences.
+     * If false we will throw an IllegalArgumentException for invalid unicode sequences.
+     */
     protected boolean _cfgLenientUnicodeEncoding;
 
     /*
@@ -1493,27 +1500,15 @@ public class CBORGenerator extends GeneratorBase
             }
             // Yup, a surrogate pair
             if (c > SURR1_LAST) { // must be from first range; second won't do
-                if (_cfgLenientUnicodeEncoding) {
-                    c = REPLACEMENT_CHAR;
-                } else {
-                    _throwIllegalSurrogate(c);
-                }
+                c = _illegalSurrogateFound(c);
             }
             // ... meaning it must have a pair
             else if (i >= len) {
-               if (_cfgLenientUnicodeEncoding) {
-                    c = REPLACEMENT_CHAR;
-                } else {
-                    _throwIllegalSurrogate(c);
-                }
+                c = _illegalSurrogateFound(c);
             }
             // ... verify that the next character is in range
             else if (str[i] < SURR2_FIRST || str[i] > SURR2_LAST) {
-                if (_cfgLenientUnicodeEncoding) {
-                    c = REPLACEMENT_CHAR;
-                } else {
-                    _throwIllegalSurrogatePair(c, str[i]);
-                }
+                c = _illegalSurrogatePairFound(c, str[i]);
             }
             // ... we have a valid surrogate pair
             else {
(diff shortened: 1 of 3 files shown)


cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -61,7 +66,14 @@
...
          * written at the beginning of a new document.
          */
-        WRITE_TYPE_HEADER(false)
+        WRITE_TYPE_HEADER(false),
+
+        /**
+         * Feature that determines if an invalid surrogate encoding found in the
+         * incoming String should fail with an exception or silently be outputed
+         * as the Unicode 'REPLACEMENT CHARACTER' (U+FFFD)
+         */
```
cowtowncoder, line 75: Could you add `@since` annotation (but see discussion on version to include in)

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -138,6 +150,8 @@ public int getMask() {
 
     protected boolean _cfgMinimalInts;
 
+    protected boolean _cfgLenientUnicodeEncoding;
```
cowtowncoder, line 153: Brief javadoc would be nice (I know, some other entries do not have it yet but... )

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -1452,20 +1425,43 @@ private final int _encode2(int i, int outputPtr, String str, int len,
             }
             // Yup, a surrogate pair
             if (c > SURR1_LAST) { // must be from first range; second won't do
-                _throwIllegalSurrogate(c);
+                if (_cfgLenientUnicodeEncoding) {
```
cowtowncoder, line 1428: One small suggestion: change `_throwIllegalSurrogate` to contain this check, and either throw exception or return surrogate value. That way rarely needed logic contained in that method, bit more compact on call site.

cowtowncoder:
Ok, first of all, thank you for the contribution! I am bit concerned about the fact that some code is trying to output invalid Unicode content, essentially, but as long as that is documented and user has to explicitly enable said feature I think that is fine.

I added some small notes in PR itself, but 2 bigger questions:

1. Since "master" is for 3.0, which might be far out, you may want to instead make PR against 2.12. I can handle merging it forward to master (wrt API changes)
2. I'll need to have a look but my main concern is with performance: since this is an edge case, it should have no measurable effect on good case (i.e. content does not have invalid surrogate characters). I can probably test it myself but thought I'll mention it -- I think changes do affect main loops.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -1356,81 +1377,33 @@ private final int _encode(int outputPtr, char[] str, int i, int end) {
...
         } while (++i < end);
         return outputPtr - outputStart;
     }
 
-    /**
-     * Helper method called when the whole character sequence is known to fit in
-     * the output buffer, but not all characters are single-byte (ASCII)
-     * characters.
-     */
-    private final int _shortUTF8Encode2(char[] str, int i, int end,
```
cowtowncoder, line 1371: Curious as to why this was removed? Or did it just get moved and diff is confused.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -1356,81 +1377,33 @@ private final int _encode(int outputPtr, char[] str, int i, int end) {
...
-
     private final int _encode(int outputPtr, String str, int len) {
         final byte[] outBuf = _outputBuffer;
         final int outputStart = outputPtr;
 
         for (int i = 0; i < len; ++i) {
             int c = str.charAt(i);
             if (c > 0x7F) {
-                return _encode2(i, outputPtr, str, len, outputStart);
+                return _encode2(i, outputPtr, str.toCharArray(), len, outputStart);
```
cowtowncoder, line 1394: So this change was made because.... ?

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
guillaumebort: It has been removed because this code was duplicated: `_shortUTF8Encode2` and `_encode2` were basically the same with the difference that one was taking a `String` as an argument and the other one was taking a `char[]`. Since this code contains some really non trivial logic I thought it would be better to not duplicate it.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
guillaumebort: I have changed the signature of `_encode2` to take a char array instead so I could replace usage of `_shortUTF8Encode2` by `_encode2` and kill `_shortUTF8Encode2`.

guillaumebort:
> 1. Since "master" is for 3.0, which might be far out, you may want to instead make PR against 2.12. I can handle merging it forward to master (wrt API changes)

Ok fair enough I will base the PR on 2.12

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
guillaumebort: Done

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
guillaumebort: Done

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
guillaumebort: Yes good idea, done.

cowtowncoder:
@guillaumebort Apologies for slow follow-up here: now back to getting this merge for 2.12.
One thing I'd need, if I hadn't yet asked would be CLA. It's here:

and usually easiest to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Only needs to be done once before the first contribution. Apologies if we already got one and I somehow missed it.

guillaumebort:
> One thing I'd need, if I hadn't yet asked would be CLA. It's here:

Thanks! I need to handle that with the legal team at my company and I do it ASAP.

cowtowncoder:
Sounds good -- we have both individual CLA that I linked earlier (and used by most contributors), as well as Corporate CLA (CCLA), at

https://github.com/FasterXML/jackson/blob/master/contributor-agreement-corporate.txt

if that makes more sense.

guillaumebort:
Should be ok now, my company (Datadog) sent a a signed copy of the Corporate CLA over.

cowtowncoder:
@guillaumebort For some reason I don't see one yet? I assume it'd be sent to `info@fasterxml.com`?

cowtowncoder:
Received the CLA.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
cowtowncoder: Ok, so this would add another allocation (str.toCharArray()). I think I'll undo this change & manually merge other changes.

cowtowncoder:
Ended up merging this manually, hence closing PR, feature is in, tests, will be included in `2.12.0-rc2`.
Thank you again for contributing this! Might make sense to add similar support for Smile, and perhaps other backends too.

Issue #228, opened 2020-11-25
Add support for decoding unassigned "simple values" (type 7)


davidlepilote:
As per the RFC-7049, it is possible to use the Major Type 7 (0b111) to pass values with no content : 
https://tools.ietf.org/html/rfc7049#section-2.3

However, passing one of these value in a CBOR mapper ends with an exception. Here is a Kotlin sample to reproduce the exception : 

```
fun main(args: Array<String>) {
    val cborMapper = ObjectMapper(CBORFactory())
    cborMapper.readValue(byteArrayOf(0b1110, 0b0001), object : TypeReference<Any>() {})
}
```

This throws the following exception : 
`Exception in thread "main" com.fasterxml.jackson.core.JsonParseException: Invalid CBOR value token (first byte): 0xe1`

It should be possible to use these kinds of values, is there a possibility to solve this problem ?

Thanks

cowtowncoder:
Thank you for reporting this: I'll take a look soon to see what is going on.

One quick question: is the `byteArrayOf()` here producing one or two bytes? At first I thought it must be two, but based on other comments I guess it should be a single byte with value of `0xE1`?

davidlepilote:
Hello,

Thank you for your prompt answer. You're completely right, I should have written `byteArrayOf(0xE1.toByte())`. The idea here is indeed to produce one byte of value `0xE1`.

I started to write some code but it doesn't work as expected. I wanted to change how the CBORParser class handles tokens of Major Type 7 but didn't manage to output the right values.

cowtowncoder:
@davidlepilote Ok. Looks like there are undefined "simple values", meaning of which is specified somewhere else. Decoding part would not be too hard, but runs into one very specific conceptual problem -- how should these be exposed as `JsonToken`? It is not possible to simply skip them (for Objects it would require hiding the whole entry; for Arrays size would differ), and without knowing expected type cannot quite map them either.

I guess the obvious answer would be `JsonToken.VALUE_EMBEDDED_OBJECT` which is already used for format-specific types (most commonly binary data); and we could perhaps add simple `CBORSimpleValue(int)` container as value.
This would also be needed on generation side, probably via `JsonGenerator.writeEmbeddedObject(Object)`, which could expect this container.

But one thing that I do not quite understand is this: I do not see any definitions for simple values beyond standard-specified ones, on:

https://en.wikipedia.org/wiki/CBOR#Simple_value

(in contrast to type tags, for which registrations exist)

[...]

PR #229, opened 2020-11-25
Add support for simple values with no content for Major Type 7

davidlepilote:
A pull request for the issue https://github.com/FasterXML/jackson-dataformats-binary/issues/228


davidlepilote:
Still some problems with the parsing

Issue #149, opened 2018-10-09
[Ion] Polymorphic deserialization does not work correctly


krackoder:
Polymorphic deserialization from Ion even when providing subtypes with Json annotation does not work correctly. The details are in this SO post -
https://stackoverflow.com/questions/52672038/missing-type-id-when-trying-to-resolve-subtype-jackson-ion-desrialization-even-a/

PR #231, opened 2020-11-30
Reference to #149 in CBORParser should be #124

jobarr-amzn:
#124 relates to a CBOR invalid integer value issue, whereas #149 is about polymorphic Ion serialization.


cowtowncoder:
Good catch, thanks!

Issue #225, opened 2020-11-03
Round-trip Ion serialization/deserialization breaks when @JsonTypeInfo annotation is present


pymaxion-amzn:
I have not done too much investigation into this problem, but it appears that the `@JsonTypeInfo` annotation might not be handled properly by the `IonObjectMapper` code path. I discovered this issue while trying to upgrade my Jackson dependencies from version 2.8.x to 2.10.x, which involved switching from an old, now-deprecated `JacksonIon` library onto this one.

Let's say that I have the following classes, which approximately model my real-world use case:
```
@JsonTypeInfo(use = Id.CLASS, include = As.PROPERTY, property = "@class")
public abstract class BaseClass { // empty }

@Data
public class Subclass extends BaseClass {
    private final String customField1;
    private final int customField2;
}
```

I would expect the following test to succeed, but it sadly does not:
```
@Test
  void subclass_RoundTripSerialization_IsSuccessful() throws IOException {
    IonObjectMapper ionObjectMapper = new IonObjectMapper();
    Subclass instance = new Subclass("some value", 42);
    IonValue instanceAsIon = ionObjectMapper.writeValueAsIonValue(instance);
    BaseClass roundTripInstance = ionObjectMapper.readValue(instanceAsIon, BaseClass.class);

    assertTrue(roundTripInstance instanceof Subclass);
    assertEquals(instance, roundTripInstance);
  }
```

This test succeeds with the older `JacksonIon` library, but does not succeed now. Specifically, the test throws an exception when executing `ionValueMapper.parse(instanceAsIon, BaseClass.class)`, something like:

```
com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Missing type id when trying to resolve subtype of [...]: missing type id property '@class'
```

Indeed, when looking at the serialized data, it does not match what I expect based on what I provided in the `@JsonTypeInfo` annotation:

```
// expected
{
  '@class':"my.package.Subclass",
  customField1:"some value",
  customField2:42
}

// actual
'my.package.Subclass'::{
  customField1:"some value",
  customField2:42
}
```

The exception message is accurate in that the serialized version does not have an `@class` type id property; the type id information is instead written as an Ion annotation. 

Note that although this example involves polymorphism, I wasn't able to get the test to succeed even with non-polymorphic types, for example:

```
@JsonTypeInfo(use = Id.CLASS, include = As.PROPERTY, property = "@class")
@Data
public class SimpleObj {
  private final int field;
}
[...]

PR #232, opened 2020-12-04
Allow disabling native type ids in IonMapper

jobarr-amzn:
This change makes the behavior of the Ion format more similar to
`YAMLMapper` in `jackson-dataformat-yaml`.

* Native type ids may be disabled in `Ion{Generator,Factory,Mapper}`
* IonParser capability introspection `canReadTypeId => true`
* IonGenerator#writeTypePrefix override removed, enabling non-native

Together these changes allow `AsPropertyDeserializer` with an
`IonParser` backing to deserialize either style while the `IonGenerator`
will honor its constructed feature set.

This should address #149 and #225.

Open questions:
* Are feature flags the preferred way to do this, or booleans like `protected boolean _cfgCreateBinaryWriters = false;
` in `IonFactory`?
* If feature flags, then should `Feature` implement `JacksonFeature` or `FormatFeature`?
* I copied the established pattern for Feature flags, but this seems like a pretty ideal use case for `EnumSet`. Thoughts?

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonGenerator.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonGenerator.java
@@ -22,17 +22,17 @@ import java.math.BigInteger;
 import java.util.Calendar;
 
 import com.fasterxml.jackson.core.Base64Variant;
+import com.fasterxml.jackson.core.FormatFeature;
 import com.fasterxml.jackson.core.JsonGenerationException;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.ObjectCodec;
 import com.fasterxml.jackson.core.StreamWriteCapability;
 import com.fasterxml.jackson.core.Version;
 import com.fasterxml.jackson.core.base.GeneratorBase;
 import com.fasterxml.jackson.core.io.IOContext;
 import com.fasterxml.jackson.core.json.JsonWriteContext;
-import com.fasterxml.jackson.core.type.WritableTypeId;
+import com.fasterxml.jackson.core.util.JacksonFeature;
 import com.fasterxml.jackson.core.util.JacksonFeatureSet;
 import com.fasterxml.jackson.dataformat.ion.polymorphism.IonAnnotationTypeSerializer;
 
@@ -48,11 +48,61 @@ import com.amazon.ion.Timestamp;
 public class IonGenerator
     extends GeneratorBase
 {
+    /**
+     * Enumeration that defines all toggleable features for Ion generators
+     */
+    public enum Feature implements FormatFeature // since 2.12
+    {
+        /**
+         * Whether to use Ion native Type Id construct for indicating type (true);
+         * or "generic" type property (false) when writing. Former works better for
+         * systems that are Ion-centric; latter may be better choice for interoperability,
+         * when converting between formats or accepting other formats. Enabled by default
+         * for backwards compatibility as that has been the behavior of
+         * `jackson-dataformat-ion` since 2.9.
+         *
+         * @see <a href="https://amzn.github.io/ion-docs/docs/spec.html#annot>The Ion Specification</a>
+         * @since 2.12
+         */
+        USE_NATIVE_TYPE_ID(true),
+        ;
+
+        protected final boolean _defaultState;
+        protected final int _mask;
+
+        /**
+         * Method that calculates bit set (flags) of all features that
+         * are enabled by default.
+         */
+        public static int collectDefaults()
+        {
+            int flags = 0;
+            for (Feature f : values()) {
+                if (f.enabledByDefault()) {
+                    flags |= f.getMask();
+                }
+            }
+            return flags;
+        }
+
+        private Feature(boolean defaultState) {
+            _defaultState = defaultState;
+            _mask = (1 << ordinal());
+        }
+
+        @Override
+        public boolean enabledByDefault() { return _defaultState; }
+        @Override
+        public boolean enabledIn(int flags) { return (flags & _mask) != 0; }
+        @Override
+        public int getMask() { return _mask; }
+    }
+
     /*
      *****************************************************************
      * Basic configuration
      *****************************************************************
-      */  
+      */
 
     /* This is the textual or binary writer */
     protected final IonWriter _writer;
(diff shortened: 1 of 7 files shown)


cowtowncoder:
First things first:

* Features or flags? For format modules, Features for sure.
* Use `FormatFeature` for format-specific read/write features -- either `ParserType.Feature` or separate `IonReadFeature` / `IonWriteFeature`. F.ex see `JsonParser.Feature` and `JsonGenerator.Feature`)
* but if there are any per-Factory features, `JacksonFeature` (see `JsonFactory.Feature`)

As to `EnumSet` -- it'd be almost nice for `EnumSet`; if `EnumSet` was immutable I'd probably have used it originally.
Using an `int` is bit old school, but at this point it is also the way it must be used for parser/generator features (`ObjectReader` and `ObjectWriter` pass them when constructing new parser/generator) to work.

cowtowncoder:
One note: timing-wise this is pretty unfortunate: semantically, this would have to wait until 2.13 as it adds new API feature.

But depending on how safe the change is, I think I could perhaps consider adding it as "dark launch" in 2.12.x; would remain undocumented until 2.13 but available for those who know about it (and do not mind the technical slight incompatibility between patch versions).

jobarr-amzn:
> but if there are any per-Factory features, `JacksonFeature` (see `JsonFactory.Feature`)

Okay, my understanding is that that is not the case here. I think the pattern I'm following now is what is wanted, except my `Feature` enums should extend `FormatFeature`. I'll keep the same set of convenience plumbing that I see in `jackson-dataformat-yaml`.

> But depending on how safe the change is, I think I could perhaps consider adding it as "dark launch" in 2.12.x; would remain undocumented until 2.13 but available for those who know about it (and do not mind the technical slight incompatibility between patch versions).

Thanks! 👍 For safety all the existing tests continue to pass unaltered and the new ones pass too ;)

Prior to this patch trying to round-trip `Subclass` like the new unit tests do results in:

```plain
com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Missing type id when trying to resolve subtype of [simple type, class com.fasterxml.jackson.dataformat.ion.polymorphism.PolymorphicRoundTripWithSerializationAnnotations$BaseClass]: missing type id property '@class'
 at [Source: UNKNOWN; line: -1, column: -1]

	at com.fasterxml.jackson.databind.exc.InvalidTypeIdException.from(InvalidTypeIdException.java:43)
	at com.fasterxml.jackson.databind.DeserializationContext.missingTypeIdException(DeserializationContext.java:1945)
	at com.fasterxml.jackson.databind.DeserializationContext.handleMissingTypeId(DeserializationContext.java:1458)
	at com.fasterxml.jackson.databind.jsontype.impl.TypeDeserializerBase._handleMissingTypeId(TypeDeserializerBase.java:307)
	at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer._deserializeTypedUsingDefaultImpl(AsPropertyTypeDeserializer.java:174)
	at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer.deserializeTypedFromObject(AsPropertyTypeDeserializer.java:113)
	at com.fasterxml.jackson.databind.deser.AbstractDeserializer.deserializeWithType(AbstractDeserializer.java:263)
	at com.fasterxml.jackson.databind.deser.impl.TypeWrappedDeserializer.deserialize(TypeWrappedDeserializer.java:74)
	at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:322)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4591)
	at com.fasterxml.jackson.dataformat.ion.IonObjectMapper.readValue(IonObjectMapper.java:224)
	at com.fasterxml.jackson.dataformat.ion.polymorphism.PolymorphicRoundTripWithSerializationAnnotations.testNativeTypeIdsEnabledByDefault(PolymorphicRoundTripWithSerializationAnnotations.java:42)
	<SNIP>...
```

I don't imagine that anyone would be sad by a successful deserialization instead though.

jobarr-amzn:
I've updated this PR to use make `IonGenerator.Feature` implement `FormatFeature`, and also removed a couple static assertion imports to better match surrounding code. 

How's it looking from your perspective? Does it look mergeable? Is there anything else I can do that would be helpful or customary?

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -35,14 +36,52 @@
 public class IonParser
     extends ParserMinimalBase
 {
+    /**
+     * Enumeration that defines all togglable features for Ion parsers.
+     */
+    public enum Feature implements FormatFeature // in 2.9
```
cowtowncoder, line 42: Change to 2.12 -- also, it is not 100% required to add this since there are no features yet -- I am ok either way, but just mention it.

cowtowncoder:
Looks good and I trust you know what you are doing here, esp. wrt Ion knowledge.
Should be able to merge soon; added just couple of tiny notes.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonGenerator.java
```diff
@@ -48,19 +48,74 @@
...
+     * Enumeration that defines all toggleable features for Ion generators
+     */
+    public enum Feature implements FormatFeature // since 2.12
+    {
+        /**
+         * Whether to use Ion native Type Id construct for indicating type (true);
+         * or "generic" type property (false). Former works better for systems that
+         * are Ion-centric; latter may be better choice for interoperability, when
+         * converting between formats or accepting other formats.
+         *
```
cowtowncoder, line 61: May want to mention that this affects writing side (whether to write native type id  or not).

But one thing that should be added is mention of default setting (something like "Enabled by default for backwards-compatibility"?).

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
jobarr-amzn: Done, nice catch 👍

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonGenerator.java
jobarr-amzn: Okay, done!

cowtowncoder:
Merged: had fun time with 3.0 merge, got it done but... somehow looks like 2 of new tests fail, so could probably need help there.

cowtowncoder:
Never mind, I had managed to accidentally remove a capability introspection method from IonParser; now works.

PR #241, opened 2021-02-09
Respect WRITE_ENUMS_USING_TO_STRING in EnumAsIonSymbolSerializer

jhhladky:
We don't need to make any changes for the corresponding READ_ENUMS_USING_TO_STRING because the
enum deserializer is the Jackson one which already checks for the feature.

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/EnumAsIonSymbolSerializer.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/EnumAsIonSymbolSerializer.java
@@ -22,19 +22,20 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonFormatVisitorWrapper;
 import com.fasterxml.jackson.databind.ser.std.StdScalarSerializer;
 
 /**
  * Serializes enumeration members as IonSymbols.
- * 
+ *
  * Use annotation
- * 
+ *
  * <pre>
  * &#64;JsonSerialize(using=EnumAsIonSymbolSerializer.class)
  * </pre>
- * 
+ *
  * on enumeration members that should serialize at symbols (which amounts to serializing without being surrounded by
  * double-quotes)
  */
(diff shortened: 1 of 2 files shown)


ion/src/main/java/com/fasterxml/jackson/dataformat/ion/EnumAsIonSymbolSerializer.java
```diff
@@ -49,7 +50,11 @@ public EnumAsIonSymbolSerializer() {
     @Override
     public void serialize(Enum<?> value, JsonGenerator g, SerializerProvider provider) throws IOException {
         if (IonGenerator.class.isAssignableFrom(g.getClass())) {
-            ((IonGenerator) g).writeSymbol(value.name());
+             String valueString = provider.isEnabled(SerializationFeature.WRITE_ENUMS_USING_TO_STRING)
```
mcliedtke, line 53: nit: looks like there is an extra space on this line

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/EnumAsIonSymbolSerializationTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+        mapper.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);
+
+        Assert.assertEquals(
+            ION_SYSTEM.singleValue("some_value"),
+            mapper.writeValueAsIonValue(SomeEnum.SOME_VALUE));
+    }
+
+    private static IonObjectMapper newMapper()
+    {
```
mcliedtke, line 48: nit: move bracket to previous line to match rest of the file

mcliedtke:
LGTM, will look to @cowtowncoder for if we can add this to the 2.12 branch/release.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/EnumAsIonSymbolSerializer.java
jhhladky: Will fix

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/EnumAsIonSymbolSerializationTest.java
jhhladky: Will fix

cowtowncoder:
@mcliedtke As long as you are happy with this & changing in a patch, I'm fine with it.

@jhhladky Happy to merge, thanks -- just one thing first: unless I have asked for and gotten CLA, from:

I'd need that before merging this. It's a one-time thing so if you have sent one just let me know and I'll double-check; and in future it's good for other contributions.
Anyway if needed, it's usually easiest to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.

Looking forward to merging this!

cowtowncoder:
Addendum to CCLA received, so I'll proceed with this. Timing is good as I am thinking of release 2.12.2 soon now (for couple of important fixes) -- and there'll probably be longer wait for 2.12.3.

cowtowncoder:
One other note: since is implied that Enum is to be serialized as symbol this might not be relevant, but outside of Ion, Enums may also be serialized using index (specified by another `SerializationFeature` or `@JsonFormat(shape=Shape.NUMBER_INT)` override). In case that was relevant.

jhhladky:
> Addendum to CCLA received, so I'll proceed with this. Timing is good as I am thinking of release 2.12.2 soon now (for couple of important fixes) -- and there'll probably be longer wait for 2.12.3.

Great to hear thanks! Do you have a date in mind for that release? Context is we have another change we'd like to try to squeeze in if possible (Should have the PR out today or tomorrow), but that will be mostly informed by your timeline.

cowtowncoder:
@jhhladky Ok if you have something else coming, I can wait -- was thinking this weekend but that is not driven by any external reasons. So I can definitely wait until, for example, following weekend.
So for now I will plan at earliest to release on friday next week (19th), unless I hear anything else.
Feel free to reach out on your plans, I'd rather not features miss release just due to unlucky timing (or similarly rush others).

PR #242, opened 2021-02-12
Add support for generating IonSexps

jhhladky:
Add writeStartSexp and writeEndSexp methods to the IonGenerator. This requires extending the
JsonWriteContext into a new class that recognizes the sexp context.

### DIFF
--- a/ion/src/test/java/com/fasterxml/jackson/dataformat/ion/GenerateSexpTest.java
+++ b/ion/src/test/java/com/fasterxml/jackson/dataformat/ion/GenerateSexpTest.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at:
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package com.fasterxml.jackson.dataformat.ion;
+
+import com.amazon.ion.IonSexp;
+import com.amazon.ion.IonSystem;
+import com.amazon.ion.IonWriter;
+import com.amazon.ion.system.IonSystemBuilder;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * End to end test verifying we can serialize sexps
+ */
+public class GenerateSexpTest {
+
+    private IonSystem ionSystem;
+    private IonObjectMapper mapper;
+
+    @Before
+    public void setup() {
+        this.ionSystem = IonSystemBuilder.standard().build();
+        this.mapper = new IonObjectMapper(new IonFactory(null, ionSystem));
+    }
+
+    @Test
+    public void topLevel() throws IOException {
+        Assert.assertEquals(
+            ionSystem.singleValue("(foo \"bar\")"),
+            mapper.writeValueAsIonValue(new SexpObject("foo", "bar")));
+    }
+
+    @Test
+    public void inList() throws IOException {
+        Assert.assertEquals(
+            ionSystem.singleValue("[(foo \"bar\"), (baz \"qux\")]"),
+            mapper.writeValueAsIonValue(
+                Arrays.asList(new SexpObject("foo", "bar"), new SexpObject("baz", "qux"))));
+    }
+
+    @Test
+    public void inObject() throws IOException {
+        Assert.assertEquals(
+            ionSystem.singleValue("{sexpField:(foo \"bar\")}"),
+            mapper.writeValueAsIonValue(new SexpObjectContainer(new SexpObject("foo", "bar"))));
+    }
+
+    @Test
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Ok, ideally I think this change would go in 2.13 since it changes types a bit... but I assume this is for a bug fix that'd be good to get out sooner. As usual I think I'd want @jobarr-amzn (or one of other developers with knowledge) to add their ok as well.
About the only suggestion I have is that in addition to a fail case (in which it might make sense to check that exception message has message expected), it'd be good to have a passing case that exercises the new code path. Some existing tests probably do that already, but something explicit is nice to guard against regressions.

I will probably need some help merging this to `master` for 3.0: change in handling of read/write contexts will make handling easier (`GeneratorBase` does not create it), but requires some changes to types used. If so I'll add a note after merging.

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/GenerateSexpTest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * End to end test verifying we can serialize sexps
+ */
+public class GenerateSexpTest {
+    @Test
+    public void testGenerateSexp() throws IOException {
```
mcliedtke, line 35: Only other testing suggestion might be to add a test for writing the data as Ion binary.

mcliedtke (commented):
Overall this looks good for adding support for reading/writing `sexp` values.

An interesting backlog item might be a way to more conveniently treat lists as an Ion `sexp`. (something like `@JsonFormat(shape = JsonFormat.Shape.SEXP`).

cowtowncoder:
@mcliedtke  On `JsonFormat.Shape`: that is an interesting challenge -- Shapes are format-agnostic, and adding new entries is not necessarily easy. There is `Shape.NATURAL` which sometimes could work (for "intrinsic" representation of something). 3.0 adds `POJO` to support logical concept of key/value pairs, but that is probably not relevant.

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/GenerateSexpTest.java
```diff
@@ -0,0 +1,77 @@
...
+    private static class SomePojoSerializer extends JsonSerializer<SomePojo> {
+        @Override
+        public void serialize(SomePojo value, JsonGenerator jsonGenerator,
+                              SerializerProvider provider) throws IOException {
+            final IonGenerator joiGenerator = (IonGenerator) jsonGenerator;
+
+            joiGenerator.writeStartSexp();
+            joiGenerator.writeSymbol(value.getSymbolField());
+            joiGenerator.writeString(value.getStringField());
+            joiGenerator.writeEndSexp();
```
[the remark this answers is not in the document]
jhhladky: Will do

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/GenerateSexpTest.java
jhhladky: Will do

mcliedtke:
> @mcliedtke On `JsonFormat.Shape`: that is an interesting challenge -- Shapes are format-agnostic, and adding new entries is not necessarily easy. There is `Shape.NATURAL` which sometimes could work (for "intrinsic" representation of something). 3.0 adds `POJO` to support logical concept of key/value pairs, but that is probably not relevant.

Yeah, this is maybe something peculiar about Ion that it has both `list` and `sexp` which are very similar ordered-list concepts. I don't think `NATURAL` or `POJO` quite capture the idea well.

jhhladky:
> Ok, ideally I think this change would go in 2.13 since it changes types a bit... but I assume this is for a bug fix that'd be good to get out sooner. 

Thanks I appreciate the flexibility! It would be really great if we could get this into 2.12. 

> About the only suggestion I have is that in addition to a fail case (in which it might make sense to check that exception message has message expected), it'd be good to have a passing case that exercises the new code path. Some existing tests probably do that already, but something explicit is nice to guard against regressions.

Sure, I'll add a test against the message and add some more tests for `IonSexp`s embedded in objects, other `IonSexp`s, and as top-level values.

cowtowncoder:
@mcliedtke right, `NATURAL` semantic imply "default" handling of some kind, and POJO property key/value aspect (and really have more to do with Java value side than output representation).

Format-specific formatting aspects is one part of configurability that is sort of missing from Jackson: while it is now possible to more easily have format-specific read/write settings, those are global and very simple (on/off). Conversely it is also quite easy to have per-mapper-format-specific settings of any kind (via ObjectMapper.Builder, sub-classable), but those are also immutable and global (that is, have to be set before use and cannot be change on per-read or per-write basis; nor scoped to specific property).

Another possible approach is to improve support to format-specific value objects (there is `JsonGenerator.writeEmbeddedObject()` for writing, and `JsonToken.VALUE_EMBEDDED_OBJECT` for reading as well as `JsonParser.getEmbeddedObject()`). This was mostly added to support exposing binary data (where it exists natively and not as base64-embedded String) but has been used for other types. Probably the most urgent need for extension in this context are "logical types" that CBOR and Avro expose; that is, things that are physically encoded as something (String, Object, Array) but have logical structure or value type specified by encoding or schema. The challenge is that of exposing and using such mappings at streaming API level, but also used by databinding (in a way replacing general-purpose databinding).

Above are just concepts that exist and might be usable; I don't know how exactly they would apply.

mcliedtke (approved):

mcliedtke:
@cowtowncoder Thanks for the background, I'll poke around and maybe create a issue for it.

Latest changes LGTM

jhhladky:
Any blockers to merging this? Thanks!

cowtowncoder:
Just my time: Jackson is just my avocation, not job. This change will likely not merge cleanly to master and I'll need to figure that out.
This is on top of my todo list fwtw.

cowtowncoder:
Was able to merge to master; also realized that pretty-printing does not really work as is and filed a new issue (#245). Will need help resolving that one, but at least this is merged.

jhhladky:
Thanks for merging this! I really appreciate the prompt responses.

cowtowncoder:
@JacekLach np! Do you think there are other changes coming soon -- I am thinking of possibly releasing 2.12.2 over the weekend, but wanted to see if there might be something else pending.

jhhladky:
Nothing else from my end :)

PR #244, opened 2021-02-15
Add support for deserializing IonTimestamps and IonBlobs

### LINKED ISSUE #243: [Ion] IonValueDeserializer can't deserialize IonTimestamps
To reproduce:
```java
IonSystem ion = IonSystemBuilder.standard().build();
IonValueMapper ionValueMapper = new IonValueMapper(ion);

IonTimestamp timestamp = ion.newUtcTimestamp(new Date());
String serialized = ionValueMapper.writeValueAsString(timestamp);
// The next line fails with the ClassCastException
IonTimestamp restored = ionValueMapper.readValue(serialized, IonTimestamp.class);
```

When deserializing an IonTimestamp, the [IonParser returns a `com.amazon.ion.Timestamp` for `getEmbeddedObject()`](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.12/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java#L396-L400), which the IonValueDeserializer [attempts to cast to an IonValue](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.12/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java#L31).

The above links also exhibit that this same problem exists for the `IonBlob` and `IonClob` types (trying to cast a `byte[]` to an `IonValue`).

jessbrya-amzn:
Fixes #243 

When using the IonValueDeserializer, if the embeddedObject is _not_ an IonValue, wraps it in the appropriate IonValue.

I'm suggesting we merge this into 2.12, but it may be worth merging into previous versions as well.

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -159,6 +159,10 @@ public class IonParser
         return PackageVersion.VERSION;
     }
 
+    public IonSystem getIonSystem() {
+        return _system;
+    }
+
     /*
     /**********************************************************
     /* Capability introspection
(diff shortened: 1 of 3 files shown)


mcliedtke (approved):
The logic changes make sense to me to ensure Timestamp/Blobs are handled. Agree with the comment on blobs, and that the blob type is the safer choice compared to clob.

Issue #246, opened 2021-02-18
Add `IonObjectMapper.builderForBinaryWriters()` / `.builderforTextualWriters()` convenience methods


cowtowncoder:
Currently `IonFactory` has methods for specifying whether binary- or textual-output should be specified.
While this can be used via `IonObjectMapper` by specifying builder for `IonFactory`, this is bit unwieldy:

```
ObjectMapper mapper = IonObjectMapper.builder(
  IonFactory.builderForBinaryWriters().build()
)
// possibly configure and
.build();
```

and it'd make sense to add short-cuts to allow:

```
ObjectMapper mapper = IonObjectMapper.builderForBinaryWriters()
   .build();
```

Should also add Javadocs for existing `builder()` methods to indicate that the default is to use binary writers.

cowtowncoder:
As per discussion, added in 2.12.2 as it's bit of a "special" release anyway :)

PR #249, opened 2021-02-19
Add convenience methods to for constructing IonObjectMapper.Builder

mcliedtke:
This was referenced in https://github.com/FasterXML/jackson-dataformats-binary/issues/246 and it appears that it may have been resolved by accident. 

In any case, this serves to just add some convenience to the creation of an IonObjectMapper:
* Explicitly creating a builder for text or binary writers
* Allow a specified `IonSystem` at builder construction time
* Add convenience methods for setting `IonGenerator.Feature` or `IonParser.Feature`

These convenience methods respect the current defaults:
* [By default textual writers are enabled](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.12/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonFactory.java#L55-L58)
* [By default, a standard `IonSystem` is used](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.12/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonFactory.java#L80-L82) (`IonSystemBuilder.standard().build()`)

I've made this change to the `2.12` branch but this isn't urgent and can be moved to `2.13`, though it should be backwards compatible

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonObjectMapper.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonObjectMapper.java
@@ -26,8 +26,10 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
 
 import com.amazon.ion.IonDatagram;
 import com.amazon.ion.IonReader;
+import com.amazon.ion.IonSystem;
 import com.amazon.ion.IonValue;
 import com.amazon.ion.IonWriter;
+import com.amazon.ion.system.IonSystemBuilder;
 
 /**
  * Specialization of {@link ObjectMapper} that will set underlying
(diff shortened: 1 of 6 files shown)


cowtowncoder:
Agreed, this is a good improvement. Just one question: should this go against 2.13? Ideally since it's API addition, it should; although 2.12.2 is bit special with quite a few non-compliant additions (to resolve existing problems).

mcliedtke:
I am not particularly strongly opinionated one way or the other. I pretty much had the same thought that 2.12.2 is already a bit special and this doesn't seem all that different to include. But it is also purely just added convenience so there is no rush and could easily wait until 2.13

cowtowncoder:
@mcliedtke Ok. With that, I think I might as well just merge it as-is :)

cowtowncoder:
One quick note: convenience enable/disable/configure methods can not be added in 3.0 (since `TokenStreamFactory` instances are immutable). That's fine wrt 2.x however.

cowtowncoder:
Ah. Was about to comment on inability to set IonParser/Generator features... until I realized they will work, just need to do it bit different. So that's fine. :)

Issue #270, opened 2021-04-01
[Jackson-dataformat-ion - 2.12x] Polymorphic deserialization breaks when upgrading from 2.8


manaigrn-amzn:
Given the following classes:

```
    @JsonTypeInfo(
        use = JsonTypeInfo.Id.NAME,
        include = JsonTypeInfo.As.PROPERTY,
        property = "base",
        visible = true
    )
    @JsonSubTypes({
        @Type(value = Subclass.class, name = "subclass"),
    })
    static public class BaseClass {
    }

    public static class Subclass extends BaseClass {
        public String base;
    }


    public static class Container {
        public BaseClass objectWithType;
    }
```

The following ion can be deserialized using Jackson-dataformat-ion 2.8 (using the unreleased branch)

```
{
    objectWithType:type::{
        base: "subclass"
        name: "Some name"
    } 
}
```

### Current Behavior
When using Jackson-dataformat-ion 2.12.x, the following exception is thrown:
```
com.fasterxml.jackson.databind.exc.InvalidTypeIdException:
Could not resolve type id 'type' as a subtype of `com.fasterxml.jackson.dataformat.ion.polymorphism.PolymorphicTypeAnnotationsTest$BaseClass`: known type ids = [PolymorphicTypeAnnotationsTest$BaseClass, subtype] (for POJO property 'objectWithType')
```

This happens because [IonParser.canReadTypeId()](https://github.com/FasterXML/jackson-dataformats-binary/blob/master/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java#L156) always returns true, it makes [Jackson-databind](https://github.com/FasterXML/jackson-databind/blob/df8d7a15a0c0d94a57c35ac31aacef5ed10de58a/src/main/java/com/fasterxml/jackson/databind/jsontype/impl/AsPropertyTypeDeserializer.java#L68) incorrectly read the [field name token](https://amzn.github.io/ion-docs/docs/spec.html#struct) `type::` as a *native type id*.

### Expected Behavior
There should not be an exception. This could be accomplished by allowing the check for *native type ids* to be disabled in the `IonParser` while deserializing, similar to what was added in the `IonGenerator` in #232 during serialization.

Then, the IonMapper could be configured with the following:
```
IonObjectMapper ionMapper = new IonValueMapper();
ionMapper.disable(IonParser.Feature.USE_NATIVE_TYPE_ID)
```
[...]

PR #271, opened 2021-04-01
Allow disabling native type ids when deserializing for #270

manaigrn-amzn:
This allows the check for native type ids to be disabled while deserializing. I would like to get feedback on this approach.

With the following example, native type ids will not be expected when deserializing.
```
IonObjectMapper mapper = new IonObjectMapper();
mapper.disable(IonParser.Feature.USE_NATIVE_TYPE_ID);
```

The full explanation of the problem this is solving can be found in #270

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -40,6 +40,18 @@ public class IonParser
      */
     public enum Feature implements FormatFeature // in 2.12
     {
+        /**
+         * Whether to expect Ion native Type Id construct for indicating type (true);
+         * or "generic" type property (false) when deserializing.
+         *<p>
+         * Enabled by default for backwards compatibility as that has been the behavior
+         * of `jackson-dataformat-ion` since 2.9 (first official public version)
+         *
+         * @see <a href="https://amzn.github.io/ion-docs/docs/spec.html#annot">The Ion Specification</a>
+         *
+         * @since 2.12.3
+         */
+        USE_NATIVE_TYPE_ID(true),
         ;
 
         final boolean _defaultState;
@@ -113,6 +125,13 @@ public class IonParser
      */
     protected JsonToken _valueToken;
 
+    /**
+     * Bit flag composed of bits that indicate which
+     * {@link IonParser.Feature}s
+     * are enabled.
+     */
+    protected int _formatFeatures;
+
     /*
     /*****************************************************************
     /* Construction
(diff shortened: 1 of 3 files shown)


ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -40,6 +40,18 @@
...
+        /**
+         * Whether to expect Ion native Type Id construct for indicating type (true);
+         * or "generic" type property (false) when deserializing.
+         *<p>
+         * Enabled by default for backwards compatibility as that has been the behavior
+         * of `jackson-dataformat-ion` since 2.9 (first official public version)
+         *
+         * @see <a href="https://amzn.github.io/ion-docs/docs/spec.html#annot">The Ion Specification</a>
+         *
+         * @since 2.12
```
cowtowncoder, line 52: If added at this point, must be marked as since 2.12.3 (but I'll add another note on this)

cowtowncoder:
Sounds reasonable. The only concern I have is that ideally new functionality should be in a new minor version (2.13).
Ion format has had a few challenges and this has not really been enforced, so it may be necessary to have one more exception here (since this may well be blocking problem), but I mention this just for future reference.
(2.13.0 release is still out 2 months or more, no release candidate yet planned).

Would like to have another opinion from other contributors, but makes sense to me. I'll add this to my todo list to re-review and make sure it can be merge before 2.12.3 is finalized (there are a few other changes and I plan to release that within 2 weeks or so).

cowtowncoder:
@manaigrn-amzn quick note: I assume this is under amazon ccla, so no need for separate cla.

manaigrn-amzn:
> [...] so it may be necessary to have one more exception here (since this may well be blocking problem) [...]

Yeah, it would be greatly appreciated to get this change released as fast as possible. We are on a tight deadline to migrate to 2.12.x, and we are blocked without this change.

> @manaigrn-amzn quick note: I assume this is under amazon ccla, so no need for separate cla.

Yes, that is correct.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
manaigrn-amzn: Got it. I will update it.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -133,15 +152,16 @@ public IonParser(IonReader r, IOContext ctxt)
      */
     @Deprecated
     public IonParser(IonReader r, IOContext ctxt, ObjectCodec codec) {
-        this(r, IonSystemBuilder.standard().build(), ctxt, codec);
+        this(r, IonSystemBuilder.standard().build(), ctxt, codec, IonFactory.DEFAULT_ION_PARSER_FEATURE_FLAGS);
     }
 
-    IonParser(IonReader r, IonSystem system, IOContext ctxt, ObjectCodec codec) {
```
cowtowncoder, line 139: We probably should retain this intermediate constructor (at least for 2.12.x); will add a deprecated variant after merge.

PR #282, opened 2021-05-22
[Avro] Fix unit test for untyped map schema generation.

### LINKED ISSUE #281: [Avro] Unit test for untyped map schema fails in 2.13 branch
Unit test `TestSimpleGeneration#testSchemaForUntypedMap` fails in `2.13` branch.

`UnsupportedOperationException` exception is thrown instead of expected `InvalidDefinitionException` exception.

```
[ERROR] Errors:
[ERROR]   TestSimpleGeneration.testSchemaForUntypedMap:166 » UnsupportedOperation Maps w...
```

Stack trace:
```
java.lang.UnsupportedOperationException: Maps with non-stringable keys are not supported yet

	at com.fasterxml.jackson.dataformat.avro.schema.MapVisitor.builtAvroSchema(MapVisitor.java:40)
	at com.fasterxml.jackson.dataformat.avro.schema.VisitorFormatWrapperImpl.getAvroSchema(VisitorFormatWrapperImpl.java:68)
	at com.fasterxml.jackson.dataformat.avro.schema.AvroSchemaGenerator.getGeneratedSchema(AvroSchemaGenerator.java:20)
	at com.fasterxml.jackson.dataformat.avro.AvroMapper.schemaFor(AvroMapper.java:187)
	at com.fasterxml.jackson.dataformat.avro.schema.TestSimpleGeneration.testSchemaForUntypedMap(TestSimpleGeneration.java:166)
```
In **2.13** `UnknownSerializer` calls `visitor.expectObjectFormat(typeHint)` whereas in **2.12**, `UnknownSerializer` calls `visitor.expectAnyFormat(typeHint)`.

It is since https://github.com/FasterXML/jackson-databind/commit/f19e26eb598349c3332ab37506d4a5db5aa90294

I think the whole fix is to update exception evaluation.

MichalFoksa:
Since 2.13, `UnsupportedOperationException` is thrown instead of `InvalidDefinitionException`.

Fixes #281

### DIFF
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/schema/TestSimpleGeneration.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/schema/TestSimpleGeneration.java
@@ -18,9 +18,9 @@ public class TestSimpleGeneration extends AvroTestBase
     {
         @JsonAlias({"nm", "Name"})
         public String name;
-        
+
         public int value;
-        
+
         List<String> other;
     }
 
@@ -51,7 +51,7 @@ public class TestSimpleGeneration extends AvroTestBase
             }
         }
     }
-    
+
     /*
     /**********************************************************
     /* Tests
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Thank you for digging into this! Ideally, I think, we should try to avoid exposing non-Jackson `RuntimeException`s, so it'd be good to see if this could be changed. This because users are less likely to catch these exceptions (wrt jackson's declared ones; type of which will change in 3.x to become unchecked but still with common `JacksonException` base type).

Not sure if exception 2.12 uses is optimal, but it seems better than `UnsupportedOperationException`.

I'll merge this first to stop failing CI and see if I can find time to further improve handling.

PR #283, opened 2021-06-08
[Avro] logicalType support for some `java.time` types.

MichalFoksa:
Functionality to support Avro schema with `logicaltype` for few java.time types. It is achieved by new `DateTimeVisitor`, new module `AvroJavaTimeModule` and couple of serializers and deserializers.

`AvroJavaTimeModule` works in milliseconds precision.

```
AvroMapper.builder()
    .addModule(new AvroJavaTimeModule())
    .build();    
```

#### Supported java types:

- `java.time.OffsetDateTime`
- `java.time.ZonedDateTime`
- `java.time.Instant`
- `java.time.LocalDate`
- `java.time.LocalTime`
- `java.time.LocalDateTime`

### Risk
I think risk comes for backwards compatibility. 

In `2.11` these Java types were somehow supported, they have got serialized either into array. If JSR310 `JavaTimeModule` with some combination of `WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS` it coudd have been serialize also into string or long.

Since `2.12` an exception is throw on serialization of any of these types on out of the box configuration, `AvroMapper.builder().build()`. (I did not investigate further what might be idea behind this new behavior.)

Exception after OffsetDateTime serialization in 2.12:
```
InvalidDefinitionException: "Any" type (usually for `java.lang.Object`) not supported: `expectAnyFormat` 
called with type [simple type, class java.time.OffsetDateTime]`
```

`DateTimeVisitor` cannot be turned off. Anyone already using JSR310 `JavaTimeModule` and disabling `WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS` will suddenly receive `logicalType` instead of `java-class` in Avro schema.

java.time.OffsetDateTime schema before PR:
````
:
{
  "type" : "long",
  "java-class" : "java.time.OffsetDateTime"
}
````

java.time.OffsetDateTime schema after PR:
````java.time.OffsetDateTime schema:
{
  "type" : "long",
  "logicalType" : "timestamp-millis"
}
````

It could be a problem for someone who already dealt with java.time types and found this workaround, but that somebody can adapt to in my opinion more Avro compatible solution. 

#### Precision
Avro supports milliseconds and microseconds previsions for date and time related logicalType(s). I have implemented milliseconds precision only (based on my needs).

It is easy to change PR to microseconds prevision only. To support both previsions I think a new `Feature` switch have to be created  - also not a problem.

#### java.util.Date
I have decided to remove `java.util.Date` from supported types due to potential backwards compatibility issues. At the moment schema for `java.util.Date` looks like this:
```
{
  "type" : "long",
  "java-class" : "java.util.Date"
}
```

After it would look like:
```
{
  "type" : "long",
  "logicalType" : "timestamp-millis"
}
```

I feel like it could cause issue to more people than issue with java.time types.

### DIFF
--- a/avro/README.md
+++ b/avro/README.md
@@ -139,7 +139,7 @@ Supported java.time types with Avro schema.
 
 #### Precision
 
-Avro supports milliseconds and microseconds previsions for date and time related logicalType(s). Only the milliseconds precision is supported.
+Avro supports milliseconds and microseconds precision for date and time related LogicalTypes, but this module only supports millisecond precision.
 
 ## Generating Avro Schema from POJO definition
 
--- a/avro/README.md
+++ b/avro/README.md
@@ -116,7 +116,7 @@ Serialization and deserialization support for limited set of `java.time` classes
 
 ```java
 AvroMapper mapper = AvroMapper.builder()
-    .addModules(new AvroJavaTimeModule())
+    .addModule(new AvroJavaTimeModule())
     .build();
 ```
  
--- a/avro/README.md
+++ b/avro/README.md
@@ -114,6 +114,10 @@ and that's about it, for now.
 ## Java Time Support
 Serialization and deserialization support for limited set of `java.time` classes to Avro with [logical type](http://avro.apache.org/docs/current/spec.html#Logical+Types) is provided by `AvroJavaTimeModule`.
 
+This module is to be used either:
+- Instead of Java 8 date/time module (`com.fasterxml.jackson.datatype.jsr310.JavaTimeModule`) or
+- to override Java 8 date/time module and for that, module must be registered AFTER Java 8 date/time module (last registration wins).
+
 ```java
 AvroMapper mapper = AvroMapper.builder()
     .addModule(new AvroJavaTimeModule())
@@ -121,7 +125,7 @@ AvroMapper mapper = AvroMapper.builder()
 ```
  
 #### Note
-Please note that time zone information is at serialization. Serialized values represent point in time, 
+Please note that time zone information is lost at serialization. Serialized values represent point in time, 
 independent of a particular time zone or calendar. Upon reading a value back time instant is reconstructed but not the original time zone.
 
 #### Supported java.time types:
--- a/avro/README.md
+++ b/avro/README.md
@@ -111,6 +111,36 @@ byte[] avroData = mapper.writer(schema)
 
 and that's about it, for now.
 
+## Java Time Support
+Serialization and deserialization support for limited set of `java.time` classes to Avro with [logical type](http://avro.apache.org/docs/current/spec.html#Logical+Types) is provided by `AvroJavaTimeModule`.
+
+```java
+AvroMapper mapper = AvroMapper.builder()
+    .addModules(new AvroJavaTimeModule())
+    .build();
+```
+ 
+#### Note
+Please note that time zone information is at serialization. Serialized values represent point in time, 
+independent of a particular time zone or calendar. Upon reading a value back time instant is reconstructed but not the original time zone.
+
+#### Supported java.time types:
+
+Supported java.time types with Avro schema.  
+
+| Type                           | Avro schema
+| ------------------------------ | -------------
+| `java.time.OffsetDateTime`     | `{"type": "long", "logicalType": "timestamp-millis"}`
+| `java.time.ZonedDateTime`      | `{"type": "long", "logicalType": "timestamp-millis"}`
+| `java.time.Instant`            | `{"type": "long", "logicalType": "timestamp-millis"}`
+| `java.time.LocalDate`          | `{"type": "int",  "logicalType": "date"}`
+| `java.time.LocalTime`          | `{"type": "int",  "logicalType": "time-millis"}`
+| `java.time.LocalDateTime`      | `{"type": "long", "logicalType": "local-timestamp-millis"}`
+
+#### Precision
+
+Avro supports milliseconds and microseconds previsions for date and time related logicalType(s). Only the milliseconds precision is supported.
+
 ## Generating Avro Schema from POJO definition
 
 Ok but wait -- you do not have to START with an Avro Schema. This module can
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
@@ -1,8 +1,7 @@
 package com.fasterxml.jackson.dataformat.avro.jsr310;
 
-import com.fasterxml.jackson.core.Version;
-import com.fasterxml.jackson.core.json.PackageVersion;
 import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.dataformat.avro.PackageVersion;
 import com.fasterxml.jackson.dataformat.avro.jsr310.deser.AvroInstantDeserializer;
 import com.fasterxml.jackson.dataformat.avro.jsr310.deser.AvroLocalDateDeserializer;
 import com.fasterxml.jackson.dataformat.avro.jsr310.deser.AvroLocalDateTimeDeserializer;
@@ -27,7 +26,7 @@ public class AvroJavaTimeModule extends SimpleModule {
     private static final long serialVersionUID = 1L;
 
     public AvroJavaTimeModule() {
-        super(PackageVersion.VERSION);
+        super(AvroJavaTimeModule.class.getName(), PackageVersion.VERSION);
 
         addSerializer(Instant.class, AvroInstantSerializer.INSTANT);
         addSerializer(OffsetDateTime.class, AvroInstantSerializer.OFFSET_DATE_TIME);
@@ -44,18 +43,4 @@ public class AvroJavaTimeModule extends SimpleModule {
         addDeserializer(LocalTime.class, AvroLocalTimeDeserializer.INSTANCE);
     }
 
-    @Override
-    public String getModuleName() {
-        return getClass().getName();
-    }
-
-    @Override
-    public Version version() {
-        return PackageVersion.VERSION;
-    }
-
-    @Override
-    public void setupModule(SetupContext context) {
-        super.setupModule(context);
-    }
 }
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
@@ -1,0 +1,61 @@
+package com.fasterxml.jackson.dataformat.avro.jsr310;
+
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.core.json.PackageVersion;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.dataformat.avro.jsr310.deser.AvroInstantDeserializer;
+import com.fasterxml.jackson.dataformat.avro.jsr310.deser.AvroLocalDateDeserializer;
+import com.fasterxml.jackson.dataformat.avro.jsr310.deser.AvroLocalDateTimeDeserializer;
+import com.fasterxml.jackson.dataformat.avro.jsr310.deser.AvroLocalTimeDeserializer;
+import com.fasterxml.jackson.dataformat.avro.jsr310.ser.AvroInstantSerializer;
+import com.fasterxml.jackson.dataformat.avro.jsr310.ser.AvroLocalDateSerializer;
+import com.fasterxml.jackson.dataformat.avro.jsr310.ser.AvroLocalDateTimeSerializer;
+import com.fasterxml.jackson.dataformat.avro.jsr310.ser.AvroLocalTimeSerializer;
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.ZonedDateTime;
+
+/**
+ * A module that installs a collection of serializers and deserializers for java.time classes.
+ */
+public class AvroJavaTimeModule extends SimpleModule {
+
+    private static final long serialVersionUID = 1L;
+
+    public AvroJavaTimeModule() {
+        super(PackageVersion.VERSION);
+
+        addSerializer(Instant.class, AvroInstantSerializer.INSTANT);
+        addSerializer(OffsetDateTime.class, AvroInstantSerializer.OFFSET_DATE_TIME);
+        addSerializer(ZonedDateTime.class, AvroInstantSerializer.ZONED_DATE_TIME);
+        addSerializer(LocalDateTime.class, AvroLocalDateTimeSerializer.INSTANCE);
+        addSerializer(LocalDate.class, AvroLocalDateSerializer.INSTANCE);
+        addSerializer(LocalTime.class, AvroLocalTimeSerializer.INSTANCE);
+
+        addDeserializer(Instant.class, AvroInstantDeserializer.INSTANT);
+        addDeserializer(OffsetDateTime.class, AvroInstantDeserializer.OFFSET_DATE_TIME);
+        addDeserializer(ZonedDateTime.class, AvroInstantDeserializer.ZONED_DATE_TIME);
+        addDeserializer(LocalDateTime.class, AvroLocalDateTimeDeserializer.INSTANCE);
+        addDeserializer(LocalDate.class, AvroLocalDateDeserializer.INSTANCE);
+        addDeserializer(LocalTime.class, AvroLocalTimeDeserializer.INSTANCE);
+    }
+
+    @Override
+    public String getModuleName() {
+        return getClass().getName();
+    }
+
+    @Override
+    public Version version() {
+        return PackageVersion.VERSION;
+    }
+
+    @Override
+    public void setupModule(SetupContext context) {
+        super.setupModule(context);
+    }
+}
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
@@ -20,6 +20,10 @@ import java.time.ZonedDateTime;
 
 /**
  * A module that installs a collection of serializers and deserializers for java.time classes.
+ *
+ * This module is to be used either:
+ *   - Instead of Java 8 date/time module (com.fasterxml.jackson.datatype.jsr310.JavaTimeModule) or
+ *   - to override Java 8 date/time module and for that, module must be registered AFTER Java 8 date/time module.
  */
 public class AvroJavaTimeModule extends SimpleModule {
 
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroJavaTimeDeserializerBase.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroJavaTimeDeserializerBase.java
@@ -1,0 +1,34 @@
+package com.fasterxml.jackson.dataformat.avro.jsr310.deser;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdScalarDeserializer;
+import com.fasterxml.jackson.databind.type.LogicalType;
+
+import java.io.IOException;
+import java.time.ZoneId;
+
+public abstract class AvroJavaTimeDeserializerBase<T> extends StdScalarDeserializer<T> {
+
+    protected AvroJavaTimeDeserializerBase(Class<T> supportedType) {
+        super(supportedType);
+    }
+
+    @Override
+    public LogicalType logicalType() {
+        return LogicalType.DateTime;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public T deserialize(JsonParser p, DeserializationContext context) throws IOException {
+        final ZoneId defaultZoneId = context.getTimeZone().toZoneId().normalized();
+        switch (p.getCurrentToken()) {
+            case VALUE_NUMBER_INT:
+                return fromLong(p.getLongValue(), defaultZoneId);
+        }
+        return (T) context.handleUnexpectedToken(_valueClass, p);
+    }
+
+    protected abstract T fromLong(long longValue, ZoneId defaultZoneId);
+}
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroJavaTimeDeserializerBase.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroJavaTimeDeserializerBase.java
@@ -8,6 +8,8 @@ import com.fasterxml.jackson.databind.type.LogicalType;
 import java.io.IOException;
 import java.time.ZoneId;
 
+import static com.fasterxml.jackson.core.JsonToken.VALUE_NUMBER_INT;
+
 public abstract class AvroJavaTimeDeserializerBase<T> extends StdScalarDeserializer<T> {
 
     protected AvroJavaTimeDeserializerBase(Class<T> supportedType) {
@@ -22,12 +24,12 @@ public abstract class AvroJavaTimeDeserializerBase<T> extends StdScalarDeseriali
     @SuppressWarnings("unchecked")
     @Override
     public T deserialize(JsonParser p, DeserializationContext context) throws IOException {
-        final ZoneId defaultZoneId = context.getTimeZone().toZoneId().normalized();
-        switch (p.getCurrentToken()) {
-            case VALUE_NUMBER_INT:
-                return fromLong(p.getLongValue(), defaultZoneId);
+        if (p.getCurrentToken() == VALUE_NUMBER_INT) {
+            final ZoneId defaultZoneId = context.getTimeZone().toZoneId().normalized();
+            return fromLong(p.getLongValue(), defaultZoneId);
+        } else {
+            return (T) context.handleUnexpectedToken(_valueClass, p);
         }
-        return (T) context.handleUnexpectedToken(_valueClass, p);
     }
 
     protected abstract T fromLong(long longValue, ZoneId defaultZoneId);
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroLocalDateDeserializer.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroLocalDateDeserializer.java
@@ -1,13 +1,7 @@
 package com.fasterxml.jackson.dataformat.avro.jsr310.deser;
 
-import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.DeserializationContext;
-import com.fasterxml.jackson.databind.JsonMappingException;
-import com.fasterxml.jackson.databind.deser.std.StdScalarDeserializer;
-
-import java.io.IOException;
 import java.time.LocalDate;
+import java.time.ZoneId;
 
 /**
  * Deserializer for {@link LocalDate} from and integer value.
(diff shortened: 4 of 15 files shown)

### CODE REVIEW

File: avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroLocalDateDeserializer.java [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+    @SuppressWarnings("unchecked")
+    @Override
+    public LocalDate deserialize(JsonParser p, DeserializationContext context) throws IOException, JsonProcessingException {
+        switch (p.getCurrentToken()) {
+            case VALUE_NUMBER_INT:
+                return fromLong(p.getLongValue());
+            default:
+                try {
+                    return (LocalDate) context.handleUnexpectedToken(_valueClass, p);
+                } catch (JsonMappingException e) {
```
cowtowncoder, line 38: Just remove the try-catch block altogether, no need to try to wrap anything, expected pattern is to let exception pass through if any.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroLocalDateDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static final AvroLocalDateDeserializer INSTANCE = new AvroLocalDateDeserializer();
+
+    protected AvroLocalDateDeserializer() {
+        super(LocalDate.class);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public LocalDate deserialize(JsonParser p, DeserializationContext context) throws IOException, JsonProcessingException {
```
cowtowncoder, line 31: It seems to me that the pattern for deserialization is similar if not identical to all types: if `VALUE_NUMBER_INT` is gotten, get `long` value, call `fromLong()`; otherwise call `handleUnexpectedToken()`.
So maybe extract a shared base class that does `deserialize()` and leaves `fromLong()` as `abstract.

Further, need to implement

    public LogicalType logicalType() { return LogicalType.DateTime; }

(added in 2.12)

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroInstantSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+    public void serialize(T value, JsonGenerator gen, SerializerProvider provider) throws IOException {
+        /**
+         * Number of milliseconds, independent of a particular time zone or calendar,
+         * from 1 January 1970 00:00:00.000 UTC.
+         */
+        final Instant instant = getInstant.apply(value);
+        gen.writeNumber(instant.toEpochMilli());
+    }
+
+    @Override
```
cowtowncoder, line 68: Similar to deserializer, maybe use a shared base class that can at least implement `acceptJsonFormatVisitor()`?
(not sure if `serialize()` has usable pattern, probably not worth extracting).

cowtowncoder:
Looks good; I added some suggestions.

Couple of things I would need:

1. Addition(s) to `avro/README.md` to explain usage would be good
2. Unless I have gotten CLA (apologies if I missed that), I'd need one (see https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf) -- the usual way is to print, fill & sign, scan/photo, email to `info` at fasterxml dot com

and with that I would be able to merge this PR to add neat new functionality in for 2.13!

Also I have one question: do you think there is any likelihood that some usage might be broken due to addition of date/time type detection? If there was, adding a configurability option for `AvroMapper` would make sense.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroLocalDateDeserializer.java
MichalFoksa: Fine - try-catch removed.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroLocalDateDeserializer.java
MichalFoksa: Good hint,

- I have created base deserializer `AvroJavaTimeDeserializerBase` with abstract method `fromLong(long longValue, ZoneId defaultZoneId)` method.

    However `ZoneId  defaultZoneId` is not used by deserializers to timezone-less types (2 out of 4 deserializers). 
Hope it is OK and ` context.getTimeZone().toZoneId().normalized()` does not add performance significant penalty.

- `logicalType()` added.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroInstantSerializer.java
MichalFoksa: At the moment I did not find enough commonalities to create single shared base serializer. Even `acceptJsonFormatVisitor` different. In 2 serializers it sets number type to LONG and in next 2 to INT.

Later I would like to add feature to serialize values with microseconds precision (BTW: it is milliseconds precision now), then serializes will have to have a precision flag and implement `ContextualSerializer`. Then it will be useful.

MichalFoksa:
@cowtowncoder Thank you for your time. I am sorry for later response, I had a short time off.

- `avro/README.md` is updated

- Contributor License Agreement is submitted (since today).

- Since `2.12` `java.time` types cannot be serialized anyway. I believe it will be more helpful then not.

MichalFoksa:
@cowtowncoder 
- How to submit additional commits to PR? 
- Is commit modification (branch rebase / commit squash) OK? 
- Do you squash branch at merge?

cowtowncoder:
First of all: absolutely no problem if things take time -- I take my time and this is all voluntary! Great to get contributions how much trouble it can be to get submission "just right" so I appreciate your work here.

Additional PRs: I think one possibility is to just close existing one(s), push new? Could also be that you can push to your original forked branch (I actually do not know for sure).
Commit modification fine from my end, whatever github allows.

I do try to squash merges where applicable, but I am not the tidiest git committer myself.

I did receive CLA, will check that in.

cowtowncoder:
Looks good; hoping to review again today or tomorrow, get merged.

avro/README.md
```diff
@@ -111,6 +111,36 @@ byte[] avroData = mapper.writer(schema)
 
 and that's about it, for now.
 
+## Java Time Support
+Serialization and deserialization support for limited set of `java.time` classes to Avro with [logical type](http://avro.apache.org/docs/current/spec.html#Logical+Types) is provided by `AvroJavaTimeModule`.
+
+```java
+AvroMapper mapper = AvroMapper.builder()
+    .addModules(new AvroJavaTimeModule())
```
cowtowncoder, line 119: `addModule()` (I can remove `s`, just noting)

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
```diff
@@ -0,0 +1,61 @@
+package com.fasterxml.jackson.dataformat.avro.jsr310;
+
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.core.json.PackageVersion;
```
cowtowncoder, line 4: Oh no, do not use this -- that would give module name of `jackson-core` :)

Instead you will want to use some of `com.fasterxml.jackson.dataformat.avro.PackageVersion` except... I'll add another note on details.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * A module that installs a collection of serializers and deserializers for java.time classes.
+ */
+public class AvroJavaTimeModule extends SimpleModule {
+
+    private static final long serialVersionUID = 1L;
+
+    public AvroJavaTimeModule() {
+        super(PackageVersion.VERSION);
```
cowtowncoder, line 30: Since we do not want to use "parent" module name related to Avro package, will want to pass both module name and version: latter is same as parent version. So:

   super(getClass().getName(), PackageVersion.VERSION);

I think.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
```diff
@@ -0,0 +1,61 @@
...
+
+        addDeserializer(Instant.class, AvroInstantDeserializer.INSTANT);
+        addDeserializer(OffsetDateTime.class, AvroInstantDeserializer.OFFSET_DATE_TIME);
+        addDeserializer(ZonedDateTime.class, AvroInstantDeserializer.ZONED_DATE_TIME);
+        addDeserializer(LocalDateTime.class, AvroLocalDateTimeDeserializer.INSTANCE);
+        addDeserializer(LocalDate.class, AvroLocalDateDeserializer.INSTANCE);
+        addDeserializer(LocalTime.class, AvroLocalTimeDeserializer.INSTANCE);
+    }
+
+    @Override
```
cowtowncoder, line 47: Can remove as long as we pass correct name via constructor.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
```diff
@@ -0,0 +1,61 @@
...
+    public String getModuleName() {
+        return getClass().getName();
+    }
+
+    @Override
```
cowtowncoder, line 52: Can remove.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
```diff
@@ -0,0 +1,61 @@
...
+    public Version version() {
+        return PackageVersion.VERSION;
+    }
+
+    @Override
+    public void setupModule(SetupContext context) {
```
cowtowncoder, line 58: Should not be needed either.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroLocalDateTimeSerializer.java
```diff
@@ -0,0 +1,60 @@
...
+        super(LocalDateTime.class);
+    }
+
+    @Override
+    public void serialize(LocalDateTime value, JsonGenerator gen, SerializerProvider provider) throws IOException {
+        /**
+         * Number of milliseconds in a local timezone, regardless of what specific time zone is considered local,
+         * from 1 January 1970 00:00:00.000.
+         */
+        final Instant instant = value.toInstant(ZoneOffset.ofTotalSeconds(0));
```
cowtowncoder, line 48: Does this not mean UTC, NOT local time zone? That is fine with me (and actually preferable over "local" anything), but if so comment should indicate that.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule_schemaCreationTest.java
```diff
@@ -0,0 +1,75 @@
...
+                {OffsetDateTime.class, Schema.Type.LONG, "timestamp-millis"},
+                {ZonedDateTime.class, Schema.Type.LONG, "timestamp-millis"},
+                {LocalDateTime.class, Schema.Type.LONG, "local-timestamp-millis"},
+                {LocalDate.class, Schema.Type.INT, "date"},
+                {LocalTime.class, Schema.Type.INT, "time-millis"},
+        });
+    }
+
+    @Test
+    public void testSchemaCreation() throws JsonMappingException {
```
cowtowncoder, line 52: One quick question here: since changes to schema visitor are applied regardless of module (I think?), would there be problems/changes to logic for existing users? That is, is there any possible backwards-compatibility concerns wrt creating more granular Avro schema types?
If there are, it might make sense to add configurability to allow disabling use of date/time types.

cowtowncoder (commented):
So, looks quite good; some minor issues wrt module metadata (added notes on how to change) but nothing major.

One question I have however is wrt definition/use of Local date/time types: this is a perennial challenge as users are VERY confused about what "local" means, assuming it means "their timezone" (for example). My understanding is that these are rather "abstract" date/time/datetime values which do not have concrete time point without addition time zone or offset.

But to serialize such values numerically, one either needs to use more complex notation (separate fields) OR bind them to a concrete timezone/offset. In latter case it's common to just use UTC, and I think this is what is done here?
That is fine in itself, esp. if that's what typical Avro usage (with apache lib, frameworks) is.
I am only concerned about documentation aspects, explaining how it works.

So... thought it's good to be explicit on exact behavior, and trying to make things as straight-forward as possible (which is hard; this is Date/Time handling which as a domain is complicated).

Anyway: just wanted to ask about this before merging. It's always possible add improvements, changes (doc and otherwise) but thought I'll mention these before merging just in case.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
MichalFoksa: Right `com.fasterxml.jackson.dataformat.avro.PackageVersion` used instead of one from `jackson.core`.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
MichalFoksa: Removed

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
MichalFoksa: Removed

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
MichalFoksa: Removed

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
MichalFoksa: Changed to: 
`super(AvroJavaTimeModule.class.getName(), PackageVersion.VERSION);`


I took inspiration from `IonJavaTimeModule` -> copy/pasting it.
BTW: `IonJavaTimeModule` is still using incorrect `PackageVersion`, see my comment https://github.com/FasterXML/jackson-dataformats-binary/commit/e78d78ab05b2d183b7bdf2da46ea6d3545d97475#r52273882

avro/README.md
MichalFoksa: Changed to `addModule`.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroLocalDateTimeSerializer.java
MichalFoksa: This is just a way to calculate number of milliseconds from 1/Jan/1970, not considering time zone. 

In Avro documentation, and also in JSR310 **_local_** is synonym to **_without time zone_**.

Events occurring 10 minutes after 1/Jan/1970 represented by `LocalDateTime` (1 January 1970 00:**10**:00.000) in New York or in Paris will both serialize to 10 minutes (precisely to number of milliseconds in 10 minutes).

I will think of better commenting that calculation.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule_schemaCreationTest.java
MichalFoksa: Hmmm,  is there a way how to control such switch from a module ? 
I guess not.

MichalFoksa:
> My understanding is that these are rather "abstract" date/time/datetime values which do not have concrete time point without addition time zone or offset.

This is correct and now I understand how mention of "local" might be confusing.
So in essence I will replace mentions of "local" with some "without timezone" meaning in documentation.
BTW: According to Avro specification [`logicalType` for datetime without timezone is `local-timestamp-millis`](http://avro.apache.org/docs/current/spec.html#Local+timestamp+%28millisecond+precision%29).


Next I will think of a switch to enable/disable `logicalType` functionality.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/AvroJavaTimeModule.java
cowtowncoder: Ah! Thank you, will fix that one.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroLocalDateTimeSerializer.java
dinomite: So local times are only meaningful if serialized & deserialized in the same time zone context.

dinomite (commented):
Caveat that I've worked a fair bit with Java time types, but very little with Avro.

avro/README.md
```diff
@@ -111,6 +111,36 @@ byte[] avroData = mapper.writer(schema)
...
+
+```java
+AvroMapper mapper = AvroMapper.builder()
+    .addModule(new AvroJavaTimeModule())
+    .build();
+```
+ 
+#### Note
+Please note that time zone information is at serialization. Serialized values represent point in time, 
+independent of a particular time zone or calendar. Upon reading a value back time instant is reconstructed but not the original time zone.
```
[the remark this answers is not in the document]
MichalFoksa: Yes it is correct.
They are deserialized using whatever `defaultZoneId1 is passed to AvroInstantDeserializer.fromLong()?

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroLocalDateTimeSerializer.java
MichalFoksa: Yes.
"local" type tells you something happened for example at user's time 16:37 or at 19/Jun/2021 16:37, but you do not know time zone, thus one cannot pinpoint exact moment the event occurred on universal time line but I think that is not purpose of these time types.

avro/README.md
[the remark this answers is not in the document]
dinomite: I think it would be clearer to say something like, "Note that time zone & offset information is not serialized—the serialized representation is only a point in time. For local time types (`OffsetDateTime` and `ZonedDateTime`) the time zone `defaultZoneId` in the Avro context is used for converting to & from the serialized representation. The same `defaultZoneId` *must* be used at serialization & deserialization time to obtain meaningful results."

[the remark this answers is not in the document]
dinomite: Having written that, the behavior feels weird to me. Would it be possible to store the offset/zoneId for the `OffsetDateTime` and `ZonedDateTime` types?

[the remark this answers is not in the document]
cowtowncoder: To me, zoned things absolutely should retain time zone and/or offset, and changing that to something else feels very much Wrong.

For local variants it may be necessary to do interim binding if (but only if) representation uses a fixed timepoint (like `long` for "milliseconds since 1. 1. 1970") -- but if so, it should be something really fixed like UTC and not whatever user happens to configure (because "default" zone id would otherwise have to match on writer and reader).

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroLocalDateTimeSerializer.java
cowtowncoder: To me it seems that given that it is necessary to temporarily "bind" local date/time, it should NOT rely on contextual zone at all -- as @dinomite suggest, that would result in reader and writer having to agree on timezone which seems like unnecessary potential failure mode.

Instead we really should just force UTC, unless there are strong reason to do something else.

FWTW, the default time offset for Jackson is actually UTC so for many users there would be no difference.
But ones changing that would get into non-operability abyss.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/deser/AvroJavaTimeDeserializerBase.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public T deserialize(JsonParser p, DeserializationContext context) throws IOException {
+        final ZoneId defaultZoneId = context.getTimeZone().toZoneId().normalized();
+        switch (p.getCurrentToken()) {
+            case VALUE_NUMBER_INT:
+                return fromLong(p.getLongValue(), defaultZoneId);
+        }
+        return (T) context.handleUnexpectedToken(_valueClass, p);
```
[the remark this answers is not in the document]
MichalFoksa: Bit easier to understand code

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroLocalDateTimeSerializer.java
MichalFoksa: @cowtowncoder 
> To me it seems that given that it is necessary to temporarily "bind" local date/time, it should NOT rely on contextual zone at all - as @dinomite suggest, that would result in reader and writer having to agree on timezone which seems like unnecessary potential failure mode.

Local date/time does not contain time zone, it does not rely on object mapper contextual time zone.
`value.toInstant(ZoneOffset.ofTotalSeconds(0))` is just a method how to calculate to serialize.

avro/README.md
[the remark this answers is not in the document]
MichalFoksa: @cowtowncoder 
> To me, zoned things absolutely should retain time zone and/or offset, and changing that to something else feels very much Wrong.

Avro specification does not aim to preserve time zone for non `local-XYZ` logical types. 

For correct deserialization into `OffsetDateTime` or `ZonedDateTime` reader and writer have to agree on time zone.

Support of `OffsetDateTime` and `ZonedDateTime` is here for convenience - I can drop it.


> For local variants it may be necessary to do interim binding if (but only if) representation uses a fixed timepoint (like long for "milliseconds since 1. 1. 1970") -- but if so, it should be something really fixed like UTC and not whatever user happens to configure (because "default" zone id would otherwise have to match on writer and reader).

For local variants, contextual time zone is not used at all.

avro/README.md
```diff
@@ -111,6 +111,36 @@ byte[] avroData = mapper.writer(schema)
...
+| `java.time.OffsetDateTime`     | `{"type": "long", "logicalType": "timestamp-millis"}`
+| `java.time.ZonedDateTime`      | `{"type": "long", "logicalType": "timestamp-millis"}`
+| `java.time.Instant`            | `{"type": "long", "logicalType": "timestamp-millis"}`
+| `java.time.LocalDate`          | `{"type": "int",  "logicalType": "date"}`
+| `java.time.LocalTime`          | `{"type": "int",  "logicalType": "time-millis"}`
+| `java.time.LocalDateTime`      | `{"type": "long", "logicalType": "local-timestamp-millis"}`
+
+#### Precision
+
+Avro supports milliseconds and microseconds previsions for date and time related logicalType(s). Only the milliseconds precision is supported.
```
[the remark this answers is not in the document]
MichalFoksa: Thank you for suggestion

avro/README.md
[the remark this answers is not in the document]
cowtowncoder: @MichalFoksa Ok, I think I better go over the changes once again & try to find what Avro specification says. Although handling of local/zoned types has expected semantics in Java 8, I vaguely recall Avro proscribing behvavior that seemed to differ... and I think for interoperability the letter of Avro spec should usually have precedence (even if I disagreed with how it was defined).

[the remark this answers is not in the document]
cowtowncoder: Even after reading and re-reading what Avro spec says about timestamp, regular and local, I have no idea what are those supposed to mean -- it seems nonsense to be blunt.

Not the part about physical storage itself (although why on earth are there separate milli- vs micro-second types?) but ... well, if NEITHER stores timezone information NOR is there ANY WAY to sync send/receiver zones, then... there seems to be no actual reason for 2 types. At all. I mean, timestamp in this sense can NEITHER be local (it is concrete physical time offset) NOR non-local (no time offset or time zone!). It is much like `java.util.Date` yet named in a confusing way, using 4 different but rather non-distinct types.
What a mess!

But to try to untangle the mess I guess there is only the one question of how would read and write operations handle these differently.

On writing side there probably cannot be any difference: physical timestamp is what it is. Whatever "local" timezone could be thought to be does not matter; change of zone/offset would not change that value.

On reading side timezone/offset is sort of arbitrary as well: value itself is concrete, although we can use whatever zone we might want.
I guess use of "context timezone" could make sense for "local" variant? For non-local I don't have strong opinion: either UTC or context timezone would be fine as far as I see it.

@MichalFoksa WDYT? Apologies for this taking long -- but I have some time now and will get this merged during this week :)

avro/README.md
```diff
@@ -111,6 +111,36 @@ byte[] avroData = mapper.writer(schema)
...
+Serialization and deserialization support for limited set of `java.time` classes to Avro with [logical type](http://avro.apache.org/docs/current/spec.html#Logical+Types) is provided by `AvroJavaTimeModule`.
+
+```java
+AvroMapper mapper = AvroMapper.builder()
+    .addModule(new AvroJavaTimeModule())
+    .build();
+```
+ 
+#### Note
+Please note that time zone information is at serialization. Serialized values represent point in time,
```
cowtowncoder, line 124: Should "is at serialization" instead be "is not included at serialization" (or something like that)?

avro/README.md
```diff
@@ -111,6 +111,36 @@ byte[] avroData = mapper.writer(schema)
 
 and that's about it, for now.
 
+## Java Time Support
+Serialization and deserialization support for limited set of `java.time` classes to Avro with [logical type](http://avro.apache.org/docs/current/spec.html#Logical+Types) is provided by `AvroJavaTimeModule`.
+
```
cowtowncoder, line 116: Another thing I realized: this should perhaps mention that this module is to be used either:

1. Instead of Java 8 date/time module or
2. to override Java 8 date/time module

and that for (2) module must be registered AFTER java 8 date/time module (last registration wins).

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/jsr310/ser/AvroLocalDateTimeSerializer.java
cowtowncoder: @MichalFoksa Ok, I think we agree: what I was trying to say is that for purposes of calculation, abstract date+time is handled as if it was in UTC/Z -- 0 offset. Which makes sense.

cowtowncoder:
@MichalFoksa I think I am mostly happy here; added minor notes on things to mention in README but I think I agree wrt ser/deser aspects given how Avro spec defines things (which is bit confusing wrt "local"/"non-local" timestamp but whatever).

The one last thing I noticed was your mention of allowing disabling of generation of logical type, as an option.
If you want to do that, I could first merge this PR, and let you work on that separately?
There are couple of ways to do that, f.ex:

1. Add `AvroGenerator.Feature`(s) (since schema generation is done on generation side, for Jackson 2.x at least)
2. Add new `AvroMapper.Feature` (since it's bit of cross-cutting concern)
3. Configure via `AvroModule`, instead of mapper -- ideally adding `Builder` (similar to how `AvroMapper` etc are configured)

avro/README.md
MichalFoksa: Added here and also to AvroJavaTimeModule.

MichalFoksa (commented):
@cowtowncoder  
Here you are requested improvements. 

Rather worry that sorry - I am really happy you review this and give feedback.

> allowing disabling of generation of logical type, as an option

I want to do it. Doing it in another PR would be best.

cowtowncoder:
Finally merged! Thank you @MichalFoksa and @dinomite -- will go in 2.13, although there can still be improvements.
I also realized that `BigDecimal` part of logical types is missing but nice to get in.

MichalFoksa:
@cowtowncoder 
Great :) Thank you.

MichalFoksa:
Hi @cowtowncoder 

Will look at it. Seems like nice Xmas project.

cowtowncoder:
Excellent. And merry Christmas :)

PR #284, opened 2021-06-10
Support base64 strings in `getBinaryValue()` for CBOR and Smile

hherman1:
Adds support for deserializing base64 strings as binary. Elasticsearch returns base64 strings
for binary fields, not native binary.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -1716,8 +1716,10 @@ public class CBORParser extends ParserBase
         if (_tokenIncomplete) {
             _finishToken();
         }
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Sounds reasonable, although I'd probably have to check if this is similar to what other backends do (I think `ParserBase` has something similar already).
It would be nice to also have a unit test showing simple usage and verify correct operation?

But one request: could you rebase this against `2.13`? Master branch is for Jackson 3.0 which is still some ways off, so this change would make sense for 2.13 which will be released much sooner.

cowtowncoder:
Thanks! I created a bit bigger patch to also support this for Smile, and for alternative access method, and for 2.13 branch.

hherman1:
Thanks so much! Sorry I dropped this, but glad we’ll be able to remove this code from our code base! :)

cowtowncoder:
@hherman1 No problem, was easy to add just needed to have a bit of time :)

PR #292, opened 2021-07-25
[Avro] Ease future customization by unifying place where a new visitor is created.

MichalFoksa:
To ease customization by a user I suggest to add protected method `AvroFormatVisitorWrapper` (ex `VisitorFormatWrapperImpl`) createVisitorWrapper() into `AvroFormatVisitorWrapperwhich` creates and returns a new instance of visitor wrapper. This method would be used on each place where new visitor instance is created in ArrayVisitor, MapVisitor and RecordVisitor.

    By having this in place it is easier to add / modify existing functionality by a user. He just needs to extend existing visitor wrapper and override particular method (just ask me how I know it).

    ```
    protected AvroFormatVisitorWrapper createVisitorFormatWrapperImpl() {
        return new AvroFormatVisitorWrapper(_schemas, _provider);
    }
    ```

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
@@ -19,16 +19,16 @@ public class ArrayVisitor
     implements SchemaBuilder
 {
     protected final JavaType _type;
-    
-    protected final DefinedSchemas _schemas;
+
+    protected final VisitorFormatWrapperImpl _visitorWrapper;
 
     protected Schema _elementSchema;
-    
-    public ArrayVisitor(SerializerProvider p, JavaType type, DefinedSchemas schemas)
+
+    public ArrayVisitor(SerializerProvider p, JavaType type, VisitorFormatWrapperImpl visitorWrapper)
     {
         super(p);
         _type = type;
-        _schemas = schemas;
+        _visitorWrapper = visitorWrapper;
     }
 
     @Override
@@ -53,7 +53,7 @@ public class ArrayVisitor
     public void itemsFormat(JsonFormatVisitable visitable, JavaType type)
             throws JsonMappingException
     {
-        VisitorFormatWrapperImpl wrapper = new VisitorFormatWrapperImpl(_schemas, getProvider());
+        VisitorFormatWrapperImpl wrapper = _visitorWrapper.createVisitorWrapper();
         visitable.acceptJsonFormatVisitor(wrapper, type);
         _elementSchema = wrapper.getAvroSchema();
     }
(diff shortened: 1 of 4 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
```diff
@@ -1,16 +1,15 @@
...
-import org.apache.avro.Schema;
-import org.apache.avro.Schema.Type;
-
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonArrayFormatVisitor;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonFormatTypes;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonFormatVisitable;
+import org.apache.avro.Schema;
```
cowtowncoder, line 9: Please do not reorder import statements automatically in future... I use ordering from JDK types to non-Jackson types to Jackson types on purpose.

cowtowncoder:
While I am +1 for improving naming, isn't this backwards-incompatible change, and something that can't really go in 2.13 but should wait for 3.0?

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/ArrayVisitor.java
MichalFoksa: Right - reordering is undone.

MichalFoksa:
:( OK - rename is undone

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/VisitorFormatWrapperImpl.java
```diff
@@ -49,6 +49,10 @@ public VisitorFormatWrapperImpl(DefinedSchemas schemas, SerializerProvider p) {
         _provider = p;
     }
 
+    protected VisitorFormatWrapperImpl createVisitorWrapper() {
```
cowtowncoder, line 52: This seems odd from naming perspective: type with method to create what seems like (but isn't) a copy.

Would it conceptually be more somethnig like "createChildWrapper()" or "createChildInstance()" or something?

cowtowncoder:
Ok so I think I can see why this is needed. I added a note where naming seems confusing wrt method -- I think I can see why naming of the class is unfortunate.

Once we get past that, maybe it would make sense to also file a follow-up issue about actual renaming like you suggested first. I'd have to think if it actually could go in 2.14, after all, or at least in 3.0.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/VisitorFormatWrapperImpl.java
MichalFoksa: I have chosen `createChildWrapper`. Did not find better name.

cowtowncoder:
LGTM! Will merge.

MichalFoksa:
Thanks!

Issue #295, opened 2021-09-06
`jackson-dataformat-ion` does not handle null.struct deserialization correctly


MartinGian:
Hi there,

`Jackson-dataformat-ion 2.12` is not handling null deserialization correctly. 

Here is the test:

```
import com.amazon.ion.IonSystem;
import com.amazon.ion.IonValue;
import com.amazon.ion.system.IonSystemBuilder;

@Test
public void shouldBeAbleToSerializeAndDeserializeNullStructs() throws Exception {
    // given
    IonSystem ionSystem = IonSystemBuilder.standard().build();
    IonObjectMapper mapper = new IonValueMapper(ionSystem)
            .configure(IonGenerator.Feature.USE_NATIVE_TYPE_ID, false)
            .configure(IonParser.Feature.USE_NATIVE_TYPE_ID, false);

    // when
    IonValue nullStruct = ionSystem.newNullStruct();
    IonValue ionValue = mapper.writeValueAsIonValue(nullStruct);

    // then
    IonValue parsedValue = mapper.readValue(ionValue, IonValue.class);
    assertThat(parsedValue, equalTo(nullStruct));
}

java.lang.AssertionError: 
Expected: <null.struct>
     but: was null
Expected :<null.struct>
Actual   :null
```

As you can see, it is deserializing `null.struct` into a Java null instead of deserializing it to an IonValue containing a null struct. This leads to `NullPointerException` and furthermore, IMO is losing data as a null struct is not the same as a Java null. It also loses any annotation that the `null.struct` may have, example: `IonValue nullStruct = ionSystem.singleValue("foo::null.struct");`, this is read by the mapper as a Java null as well.

The serialization works correctly, it is the deserialization that is not parsing this null values correctly.

Playing a bit with it, if you change the implementation of [IonValueDeserializer](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.13/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java) and add the following override, the test above passes.

```
@Override
public IonValue getNullValue(DeserializationContext ctxt) throws JsonMappingException {
    try {
        return deserialize(ctxt.getParser(), ctxt);
    } catch (IOException e) {
        throw new JsonMappingException(e.getMessage());
    }
}
```

It would be great if someone with more experience with Jackson can validate this. I'm happy to create a PR with the change if someone from your team confirms that this is the correct implementation.

Regards.

[...]

PR #296, opened 2021-09-08
Fix ion null deserialization

MartinGian:
Fixes the `IonValueDeserializer` to be able to deserialize serialized null values correctly. [Ion does have first class `null` values](https://amzn.github.io/ion-docs/docs/spec.html#null) so a deserialized `null` value should never end up being a Java `null` value.

There is more information about this change in https://github.com/FasterXML/jackson-dataformats-binary/issues/295

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
@@ -66,7 +66,7 @@ class IonValueDeserializer extends JsonDeserializer<IonValue> {
             }
 
             return super.getNullValue(ctxt);
-        } catch (Exception e) {
+        } catch (IOException e) {
             throw JsonMappingException.from(ctxt, e.toString());
         }
     }
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
@@ -53,4 +53,21 @@ class IonValueDeserializer extends JsonDeserializer<IonValue> {
         throw JsonMappingException.from(jp, "Cannot deserialize embedded object type "
                 + embeddedObject.getClass().getCanonicalName() + " into IonValue");
     }
+
+    @Override
+    public IonValue getNullValue(DeserializationContext ctxt) throws JsonMappingException {
+        try {
+            Object embeddedObj = ctxt.getParser().getEmbeddedObject();
+            if (embeddedObj instanceof IonValue) {
+                IonValue iv = (IonValue) embeddedObj;
+                if (iv.isNullValue()) {
+                    return iv;
+                }
+            }
+
+            return super.getNullValue(ctxt);
+        } catch (Exception e) {
+            throw JsonMappingException.from(ctxt, e.toString());
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
```diff
@@ -53,4 +53,21 @@ public IonValue deserialize(JsonParser jp, DeserializationContext ctxt) throws I
...
+            Object embeddedObj = ctxt.getParser().getEmbeddedObject();
+            if (embeddedObj instanceof IonValue) {
+                IonValue iv = (IonValue) embeddedObj;
+                if (iv.isNullValue()) {
+                    return iv;
+                }
+            }
+
+            return super.getNullValue(ctxt);
+        } catch (Exception e) {
```
mcliedtke, line 69: Can we catch the specific exception here (`IOException` I believe)?

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
```diff
@@ -0,0 +1,171 @@
...
+    public void shouldBeAbleToSerializeAndDeserializeStringData() throws Exception {
+        StringData source = new StringData();
+        source.put("a", "1");
+        source.put("b", null);
+
+        IonValue data = ION_VALUE_MAPPER.serialize(source);
+        StringData result = ION_VALUE_MAPPER.parse(data, StringData.class);
+
+        assertEquals(source, result);
+    }
```
mcliedtke, line 166: Is this a sanity test to ensure that non-IonValue deserialization is not impacted?

mcliedtke (changes requested):

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
MartinGian: Sure. I'll update.

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
MartinGian: Yes. I've included some test cases not specifically covering the change but the entire `IonValueDeserializer` class since I was adding a new test class and added those as a kind of smoke/regression test.

mcliedtke:
Changes LGTM. @cowtowncoder, do you mind merging? I think 2.13 would likely make the most sense.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
```diff
@@ -53,4 +53,21 @@ public IonValue deserialize(JsonParser jp, DeserializationContext ctxt) throws I
         throw JsonMappingException.from(jp, "Cannot deserialize embedded object type "
                 + embeddedObject.getClass().getCanonicalName() + " into IonValue");
     }
+
+    @Override
+    public IonValue getNullValue(DeserializationContext ctxt) throws JsonMappingException {
+        try {
+            Object embeddedObj = ctxt.getParser().getEmbeddedObject();
```
cowtowncoder, line 60: My spider sense is tingling... am not confident this will reliably work. But maybe it does; `instanceof` acts as `null` check.

But one thing that will definitely be needed is override of:

```
    @Override
    public AccessPattern getNullAccessPattern() {
       return AccessPattern.DYNAMIC;
    }
```

as otherwise null value may be cached (default is `AccessPattern.STATIC`) and so all further deserializations might use the cached value.

cowtowncoder:
Will merge (I have some concerns as to whether this will work reliably due to reliance on underlying `JsonParser` -- but there are tests and so perhaps it does -- and I don't have better ideas for solving it either). Unless the null value could be derived from type declaration?

But before merging, I'd suggest one addition that I mentioned, to avoid possibility of "wrong" null value getting cached.
(I forget if caching is done or not, but if it was leaving out access mode could lead to pretty weird failures).

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
MartinGian: Thanks @cowtowncoder. I updated the branch with this change.

PR #307, opened 2021-12-20
Set StreamReadCapability.EXACT_FLOATS=true

htmldoug:
This gets the ball rolling setting the `StreamReadCapability.EXACT_FLOATS` flag from https://github.com/FasterXML/jackson-core/pull/733 (issue: https://github.com/FasterXML/jackson-core/issues/730).

- [x] Smile
- [x] CBOR
- [ ] Avro
- [ ] Protobuf
- [ ] Ion

I'm pausing here to take feedback (on both PRs) before proceeding with the last three formats (which I'm slightly less familiar with anyway).

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -420,8 +420,7 @@ public class CBORParser extends ParserMinimalBase
 
     @Override // since 2.12
     public JacksonFeatureSet<StreamReadCapability> getReadCapabilities() {
-        // Defaults are fine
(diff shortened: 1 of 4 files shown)


rodtan (approved):

cowtowncoder:
As per my other note, let's split things so that we'll

1. First add new read capability, overrides for formats, basic tests to verify backend exposes values as exposed
2. Then add functionality that uses capability detection (as there are a few details to figure out, as per my notes on jackson-core issue)

cowtowncoder:
ok, merged: I can take care of the other backends too; added Avro, test.

cowtowncoder:
Clarification: I'll take care of adding capability setting, introspection, but if you have spare cycles for testing actual behavior that'd be great. If not, we'll get those eventually. :)

PR #313, opened 2022-03-04
Fix #311 by wrapping IonExceptions in IonParser

### LINKED ISSUE #311: `IonObjectMapper` does not throw JacksonException for some invalid Ion
In some cases, the IonObjectMapper is not correctly catching and wrapping IonExceptions. For example, given this class:
```java
class FooBar {
    int foo;
    List<Object> bar;
}
```
When attempting to deserialize invalid Ion, sometimes we get a JsonMappingException, and other times an IonException.
```java
IonObjectMapper mapper = IonObjectMapper.builder().build();

// throws a JsonMappingException
FooBar fb1 = mapper.readValue("{ foo: 1, bar: () ] }", FooBar.class); 
// throws an IonException
FooBar fb1 = mapper.readValue("{ foo: 1, bar: ( ] )  }", FooBar.class); 
```

Why does this matter? Some other libraries (eg. Ktor, in the `JacksonConverter`) expect particular exceptions (eg. `JsonParseException` or `JsonMappingException`) from an `ObjectMapper`, and behave in unexpected ways when the `IonException` leaks through.

I believe the fix for this is to catch and wrap `IonException` as a `JsonParseException` in the `IonParser.nextToken()` method, and I intend to work on a PR for it.

popematt:
Moves the calls to `IonReader.stringValue()`, `IonReader.getTypeAnnotations()`, `IonReader.getFieldName()`, and `IonReader.next()` into `try`/`catch` blocks so that `IonException`s can be caught and wrapped in a `JsonParseException`.

This change cannot go into 2.13 because it adds a new checked exception to `IonParser.getTypeAnnotations()`, which changes the public API. However, searching for uses of this method on GitHub doesn't reveal any uses of it aside from other places in `jackson-dataformats-binary`, so including it in 2.14 (rather than 3.0) is unlikely to cause problems for any consumers of this library.

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -265,7 +265,13 @@ public class IonParser
             case FIELD_NAME:
                 return getCurrentName();
             case VALUE_STRING:
-                return _reader.stringValue();
+                try {
+                    // stringValue() will throw an UnknownSymbolException if we're
+                    // trying to get the text for a symbol id that cannot be resolved.
+                    return _reader.stringValue();
+                } catch (UnknownSymbolException e) {
+                    throw _constructError(e.getMessage(), e);
+                }
             case VALUE_NUMBER_INT:
             case VALUE_NUMBER_FLOAT:
                 Number n = getNumberValue();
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Sounds good. I assume your contribution would be covered by Amazon CCLA? If so, let me know and I can merge this in for inclusion.

Thank you for doing the due diligence here wrt compatibility, tests, much appreciated! Really wish more contributions were this thoughtful (many are, but even more!).

popematt:
I have a personal CCLA, and I believe I am covered by the Amazon CCLA. (If I must choose one, then it's probably best to go with the Amazon CCLA.)

cowtowncoder:
Ok that's great -- I do not have to choose (there is no direct linkage between PR/issue and CLA), just want to make sure there is one on file. Apologies for forgetting that you just sent the personal one month ago. :)

Issue #317, opened 2022-03-22
IonValueDeserializer does not handle getNullValue correctly for a missing property


atokuzAmzn:
Hello,

Following https://github.com/FasterXML/jackson-dataformats-binary/issues/295, we switched to 2.13 from 2.12 to be able to handle null.struct deserialization correctly but we bumped into a different problem in 2.13

We were trying to deserialize a test input into an entity with an optional IonValue field. We received IndexOutOfBoundsException when this property was missing in the input.

This is the stacktrace of the error:
```
at ExampleEntityTest.deserialization test without optional data(ExampleEntityTest.groovy:219)
Caused by: java.lang.IndexOutOfBoundsException: 0
at com.amazon.ion.impl.lite.IonContainerLite.get_child(IonContainerLite.java:663)
at com.amazon.ion.impl.lite.IonContainerLite.get(IonContainerLite.java:151)
at com.fasterxml.jackson.dataformat.ion.IonParser.getIonValue(IonParser.java:424)
at com.fasterxml.jackson.dataformat.ion.IonParser.getEmbeddedObject(IonParser.java:442)
at com.fasterxml.jackson.dataformat.ion.ionvalue.IonValueDeserializer.getNullValue(IonValueDeserializer.java:61)
at com.fasterxml.jackson.dataformat.ion.ionvalue.IonValueDeserializer.getNullValue(IonValueDeserializer.java:32)
at com.fasterxml.jackson.databind.JsonDeserializer.getAbsentValue(JsonDeserializer.java:350)
at com.fasterxml.jackson.databind.deser.impl.PropertyValueBuffer._findMissing(PropertyValueBuffer.java:203)
at com.fasterxml.jackson.databind.deser.impl.PropertyValueBuffer.getParameters(PropertyValueBuffer.java:158)
at com.fasterxml.jackson.databind.deser.ValueInstantiator.createFromObjectWith(ValueInstantiator.java:288)
at com.fasterxml.jackson.databind.deser.impl.PropertyBasedCreator.build(PropertyBasedCreator.java:202)
at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeUsingPropertyBased(BeanDeserializer.java:518)
```
The problem occurs when IonValueDeserializer tries to execute the overridden getNullValue for the missing property. In the beginning of this method it tries to get the embedded object from IonParser.
```java
    @Override
    public IonValue getNullValue(DeserializationContext ctxt) throws JsonMappingException {
[...]

Issue #318, opened 2022-03-25
IonValueDeserializer's getNullValue doesn't deserialize java nulls correctly


atokuzAmzn:
### Problem definition
This simple test case that I have added to IonValueDeserializerTest class fails:

```java
    @Test
    public void shouldBeAbleToDeserializeJavaNullValue() throws Exception {
        IonValueData source = new IonValueData();
        source.put("c", null);

        IonValue data = ION_VALUE_MAPPER.writeValueAsIonValue(source);
        IonValueData result = ION_VALUE_MAPPER.readValue(data, IonValueData.class);

        assertEquals(source, result);
    }
```
All the Ion core types have different null values (null.struct, null.list etc) and the fix that comes with https://github.com/FasterXML/jackson-dataformats-binary/issues/295 provides proper deserialization of these different null types. But it doesn't handle java nulls correctly.

Apart from these different null values, Ion also has a raw *null* type which implements IonNull interface. Only this raw null type implements IonNull and the other null values (null.struct, null.list etc) are valid IonValues and they are NOT implementing IonNull interface.
```java
public interface IonNull extends IonValue
```
When serialized, Java nulls are translated into Ion’s raw null type and they become IonNull. And naturally when deserialized it returns this IonValue (aka IonNull) which doesn’t match with Java null and causes the test listed above to fail.

### Solution Proposal

We can fix this issue by adding an extra check  when deserializing null values. If the embedded object is actually an IonNull then it returns java null as deserialization response. Of course this will NOT affect the other type specific null values like null.struct or null.list etc. because they are not an instance of IonNull so they will continue to be deserialized into their respective type. 

Here is the proposal for getNullValue method with this extra IonNull check. (The code below also includes the fix proposal for https://github.com/FasterXML/jackson-dataformats-binary/issues/317)
```java
    @Override
    public IonValue getNullValue(final DeserializationContext ctxt) throws JsonMappingException {
        try {
            final JsonParser parser = ctxt.getParser();
            if (parser.getCurrentToken() == JsonToken.VALUE_NULL) {
                final Object embeddedObj = parser.getEmbeddedObject();
                if ((embeddedObj instanceof IonValue) && !(embeddedObj instanceof IonNull)) {
                    final IonValue iv = (IonValue) embeddedObj;
                    if (iv.isNullValue()) {
                        return iv;
                    }
                }
            }
            return super.getNullValue(ctxt);
        } catch (IOException e) {
[...]

PR #319, opened 2022-03-28
Fix getNullValue of IonValueDeserializer. 

atokuzAmzn:
This PR aims to resolve the issues https://github.com/FasterXML/jackson-dataformats-binary/issues/317 and https://github.com/FasterXML/jackson-dataformats-binary/issues/318

Changes:
1. Do not call getEmbeddedObject when current token is END_OBJECT
2. Make sure java nulls are deserialized correctly
3. Update unit test case with nulls

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
@@ -63,7 +63,8 @@ class IonValueDeserializer extends JsonDeserializer<IonValue> {
             final JsonParser parser = ctxt.getParser();
             if (parser != null && parser.getCurrentToken() != JsonToken.END_OBJECT) {
                 final Object embeddedObj = parser.getEmbeddedObject();
-                if ((embeddedObj instanceof IonValue) && !(embeddedObj instanceof IonNull)) {
+                if ((embeddedObj instanceof IonValue)
+                    && ( (parser.getTypeId() != null) || !(embeddedObj instanceof IonNull))) {
                     final IonValue iv = (IonValue) embeddedObj;
                     if (iv.isNullValue()) {
                         return iv;
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
@@ -14,6 +14,8 @@
 
 package com.fasterxml.jackson.dataformat.ion.ionvalue;
 
+import com.amazon.ion.IonNull;
+import com.fasterxml.jackson.core.JsonToken;
 import java.io.IOException;
 
 import com.fasterxml.jackson.core.JsonParser;
@@ -58,11 +60,14 @@ class IonValueDeserializer extends JsonDeserializer<IonValue> {
     @Override
     public IonValue getNullValue(DeserializationContext ctxt) throws JsonMappingException {
         try {
-            Object embeddedObj = ctxt.getParser().getEmbeddedObject();
-            if (embeddedObj instanceof IonValue) {
-                IonValue iv = (IonValue) embeddedObj;
-                if (iv.isNullValue()) {
-                    return iv;
+            final JsonParser parser = ctxt.getParser();
+            if (parser != null && parser.getCurrentToken() != JsonToken.END_OBJECT) {
+                final Object embeddedObj = parser.getEmbeddedObject();
+                if ((embeddedObj instanceof IonValue) && !(embeddedObj instanceof IonNull)) {
+                    final IonValue iv = (IonValue) embeddedObj;
+                    if (iv.isNullValue()) {
+                        return iv;
+                    }
                 }
             }
 
--- a/ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
+++ b/ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
@@ -116,17 +116,30 @@ public class IonValueDeserializerTest {
         IonValueData source = new IonValueData();
         source.put("a", null); // Serialized to IonNull, deserialized back to java null
         source.put("b", ion("null.bool"));
+        source.put("ba", ion("foo::null.bool"));
         source.put("c", ion("null.int"));
+        source.put("ca", ion("foo::null.int"));
         source.put("d", ion("null.float"));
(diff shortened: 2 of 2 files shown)


atokuzAmzn:
@cowtowncoder Can you add @popematt as reviewer?

popematt (changes requested):
The code itself is fine in that it does exactly what you say it does. However, I am concerned about the proposed solution for deserializing nulls (#318). This is technically a backwards-incompatible change that you're introducing, which means that [it can't be merged into in any 2.x branch](https://github.com/FasterXML/jackson/blob/master/CONTRIBUTING.md#user-content-backwards-compatibility).

Can you make the new behavior opt-in through some sort of configuration?

atokuzAmzn:
> This is technically a backwards-incompatible change that you're introducing, which means that [it can't be merged into in any 2.x branch](https://github.com/FasterXML/jackson/blob/master/CONTRIBUTING.md#user-content-backwards-compatibility).
> 
Actually what I am proposing is to fix backward compatibility. After https://github.com/FasterXML/jackson-dataformats-binary/pull/296, 2.13 already became incompatible.

Just take this simple test case below.
 **_This test succeeds in 2.12 but right now it fails in 2.13_**. 
If we had a test case like this in 2.12 this would have been caught.

```java
    @Test
    public void shouldBeAbleToDeserializeJavaNullValue() throws Exception {
        IonValueData source = new IonValueData();
        source.put("c", null);

        IonValue data = ION_VALUE_MAPPER.writeValueAsIonValue(source);
        IonValueData result = ION_VALUE_MAPPER.readValue(data, IonValueData.class);

        assertEquals(source, result);
    }
```

The problem defined in issue https://github.com/FasterXML/jackson-dataformats-binary/issues/295 was about `null.struct`s not being deserialized correctly. Definitely this was a miss and should have been fixed but while fixing the issue we shouldn't have broken the java null deserialization. With this proposal both `null.struct`s (etc) and java nulls will be deserialized correctly.

popematt:
@atokuzAmzn Thank you for explaining that; this makes sense to me now.

@cowtowncoder, should this be merged into the 2.13 branch as well as 2.14?

popematt (approved):
It might be useful for to add a unit test that's something like this just to make it absolutely clear how Java `null` and Ion `null`/`null.null` should interact with each other in this context.

```java
IonValueData source = new IonValueData();
source.put("a", null);
source.put("b", ion("null"));

IonValue data = ION_VALUE_MAPPER.writeValueAsIonValue(source);
IonValueData result = ION_VALUE_MAPPER.readValue(data, IonValueData.class);

IonValueData expected = new IonValueData();
expected.put("a", null);
expected.put("b", null);

assertEquals(expected, result);
```

atokuzAmzn:
@popematt I have added an extra test case

cowtowncoder:
As I said, if others (@popematt ) agree, I am ok changing it to 2.13. I am just saying that for users who might have started with 2.13.0, change in behavior in a patch might be surprising.
But I do not have full context or understanding to know if it might be that 2.13.x behavior so far is so broken there's no legitimate use case -- if so, yeah, patch it up.

Just let me know either way.

tgregg:
@atokuzAmzn  The Ion specification allows `null` to have annotations, just like any other value. E.g. `'com.foo.Bar'::null`. Under this proposal, do we lose all annotations on untyped nulls? (Note: this should be added to the unit tests.) If so, there needs to be an alternate behavior capable of preserving the annotations, which are part of the Ion data model. Should we return `IonNull` when reading annotated untyped nulls and Java `null` when reading un-annotated untyped nulls? Or should there be configuration to allow users to receive Java `null` for all Ion nulls, allowing them to opt-in to the possibility of data loss?

atokuzAmzn:
@tgregg I have added another check to implement your first option and also added a new unit test case.

mcliedtke:
To me, the 2.13 behavior seems like the right handling for Ion nulls. An `IonValue` represents the actual Ion domain data object and in Ion, the `null`'s are types and have an `IonValue` representation. An untyped Ion `null` should result in non-(java)-null `IonValue` which is of type `IonNull`. On top of that, the divergence of handling of typed nulls, untyped null, and annotated nulls  in the latest revision seem really inconsistent / confusing. Why should the unannotated, untyped null be "special" and result in a Java null?

I realize this is different from 2.12, but the 2.13 just seems more correct. I'd potentially be open to a feature flag that allows a user to specify the legacy behavior if there was a simple way.

atokuzAmzn:
This is moving into a deadlock:)

And I think it was my mistake trying to solve two issues (https://github.com/FasterXML/jackson-dataformats-binary/issues/317 and https://github.com/FasterXML/jackson-dataformats-binary/issues/318) with one PR. Because as far as I can understand most of the discussion revolves around https://github.com/FasterXML/jackson-dataformats-binary/issues/318, but I could actually create a separate PR for https://github.com/FasterXML/jackson-dataformats-binary/issues/317 to solve that separately.

So back to https://github.com/FasterXML/jackson-dataformats-binary/issues/318. I accept that the last commit made it confusing. But as far as I understand there can never be a full solution to this problem. So we have to make a trade off.

Either,
we will deserialize java nulls to IonNull
or,
we will deserialize IonNulls(not null.struct etc they are NOT IonNulls) to java null.

Whatever we choose, we will make a wrong deserialization. There is no escape from this. 
So we just need to decide which wrong we want to do by default. 

Even if we make this configurable we will still need a default value. 
So the question is, what should be the default behaviour?

And I believe the answer to this question lies in which use case occurs more often than the other. I am not an expert at all in this but I believe serializing java nulls probably occurs much more often than serializing IonNulls. And just to add as confirmation, in the internal implementation which we were using for a long time, the choice was to return java nulls for both of them.

atokuzAmzn:
Well, I don't have anything else to add on this issue, if this is the decision, then it is.

I have published a separate PR https://github.com/FasterXML/jackson-dataformats-binary/pull/320 for https://github.com/FasterXML/jackson-dataformats-binary/issues/317, which is much more straightforward. Please review when you have time.

cowtowncoder:
And just to make it clear: I am happy to merge whatever is agreed upon by maintainers, but it may be simplest for @mcliedtke to merge. The only (?) caveat is that whoever merges PRs also needs to ensure they get forward-merged to later branches as appropriate (2.13 -> 2.14 -> master).

cowtowncoder:
Now that #317 is resolved, what would be the way forward here, if any? No particular hurry but wanted to add an update if there might be progress.

tgregg:
I agree with @mcliedtke 's assessment and suggested workaround (using `JsonInclude(Include.NON_NULL)`) when Java `null` is required instead of `IonNull`. Accordingly, it looks like this can be closed without changes.

atokuzAmzn:
In the end, we needed to make a choice so I am also OK to close this issue.

PR #338, opened 2022-09-07
Overwrite writeStartObject method with size field

szysas:
Overwrite `writeStartObject(Object forValue, int size)` method in `CBORGenerator` class.


cowtowncoder:
Good catch, thank you!

Issue #339, opened 2022-09-08
Invalid Avro file produced using SequenceWriter


willsoto:
While documentation on writing Avro to a file is sparse, I have managed to piece some stuff together but I am still getting an error.

- https://groups.google.com/g/jackson-user/c/yOXaJvAMzfg

Here is some sample code:

```java
final var avroFactory = AvroFactory.builderWithApacheDecoder().enable(AvroGenerator.Feature.AVRO_FILE_OUTPUT).build();

final var generator = new AvroSchemaGenerator().enableLogicalTypes();

final var mapper = AvroMapper.builder(avroFactory).addModule(new AvroJavaTimeModule()).build();
mapper.acceptJsonFormatVisitor(Thing.class, generator);

final var avroSchema = generator.getGeneratedSchema();

final var file = Files.createTempFile("something", ".avro").toFile();

final var out = new ByteArrayOutputStream();
final var writer = mapper.writer(avroSchema).writeValues(out);

// in a loop
writer.write(thing);

// after loop
writer.close();

try (FileOutputStream outputStream = new FileOutputStream(file)) {
  out.writeTo(outputStream);
}
```

When checking the resultant file using [`avro-tools`](https://avro.apache.org/), I get the following error:

```
avro-tools tojson something.avro

22/09/08 18:36:55 WARN util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Exception in thread "main" org.apache.avro.AvroRuntimeException: java.io.IOException: Invalid sync!
	at org.apache.avro.file.DataFileStream.hasNext(DataFileStream.java:224)
	at org.apache.avro.tool.DataFileReadTool.run(DataFileReadTool.java:97)
	at org.apache.avro.tool.Main.run(Main.java:67)
	at org.apache.avro.tool.Main.main(Main.java:56)
Caused by: java.io.IOException: Invalid sync!
	at org.apache.avro.file.DataFileStream.nextRawBlock(DataFileStream.java:319)
	at org.apache.avro.file.DataFileStream.hasNext(DataFileStream.java:213)
	... 3 mor
```

According to some searching, the `Invalid sync!` error occurs when the file hasn't been stitched together properly, but it's unclear to me what I need to do in code to get that to happen. I've looked through most of the Avro tests in this repo and I cannot find one that actually writes to a file and then de-serializes from that file. 

[...]

PR #340, opened 2022-09-10
Add failing test case for issue #339

willsoto:
See #339 for more info

### DIFF
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/failing/FileSerializationTest.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/failing/FileSerializationTest.java
@@ -1,0 +1,66 @@
+package com.fasterxml.jackson.dataformat.avro.failing;
+
+import com.fasterxml.jackson.databind.SequenceWriter;
+import com.fasterxml.jackson.dataformat.avro.AvroFactory;
+import com.fasterxml.jackson.dataformat.avro.AvroGenerator;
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Hmmh. As per my comment on #339, I am NOT 100% sure this is a valid test. In particular, it will not add any sync or file header beyond initial header since the output is considered a single logical stream.
But I guess I can just merge it and we can determine validity over time, at least there is a reproduction.

willsoto:
I'm more than happy to tweak the test case as necessary.

I wasn't quite sure what your intention was with that comment - I thought you may have just been positing one way it _could_ work.

cowtowncoder:
@willsoto Right, my comment was bit ambiguous. My guess is that this has no chance of working as things are; but what I am not sure of is as to whether this should be the way to use API.  To know that I'd need to spend some time again reading Avro spec wrt file encoding. Specifically about whether Avro has a "Stream" concept distinct from Arrays of things. If so, it would conceptually match Jackson's `SequenceWriter`.
And from that, whether there is additional sync/marker requirement between elements -- it looks like there is, judging by exception message. That would require solving; Jackson does actually have the concept of "Root Value Separator"s which _might_ work. Not exactly as-is, since separator is assumed to be `String`. But... maybe with some tweaks.

Since I don't quite know if this is valid or not, I'd be happy to include it under `failing/` for storage, being easier to find whenever someone (myself or something else) has time to dig in and see if there's a solution.

willsoto:
@cowtowncoder CLA sent.

PR #344, opened 2022-11-18
use String constructor directly for short ascii values

### LINKED ISSUE #342: Possible performance improvement on jdk9+
We have a use-case for smile that involves parsing a lot of short ASCII string values and [_decodeShortAsciiValue] shows up prominently in profiles. This implementation copies the input bytes into a character array which is then used to create the String. On newer Java versions there is a significant improvement by just using the String constructor that takes a byte array ([sample change]). This is because In jdk9 and later the internal representation of `String` changed ([JEP 254](https://openjdk.org/jeps/254)) and creating from an ASCII byte array can basically just do `Arrays.copyOfRange`. Unfortunately, this approach seems to be significantly slower on jdk8.

Results from some quick benchmarks using JMH:

```
Benchmark                       Mode  Cnt       Score       Error   Units

java 8 baseline                thrpt    5  344448.375 ± 46864.690   ops/s
java 8 with change             thrpt    5  194250.482 ±  7374.829   ops/s

java 17 baseline               thrpt    5  301991.539 ± 52160.837   ops/s
java 17 with change            thrpt    5  413394.519 ± 16857.477   ops/s
```

I'm assuming it would be unacceptable to take that hit on java 8 at this time. Are there any existing examples in jackson where behavior is conditional on the java version used? 

[_decodeShortAsciiValue]: https://github.com/FasterXML/jackson-dataformats-binary/blob/b082f58968951967195344ec99ab1ce22327d7d5/smile/src/main/java/com/fasterxml/jackson/dataformat/smile/SmileParser.java#L2396-L2431
[sample change]: https://github.com/brharrington/jackson-dataformats-binary/commit/65e8a1e3ed3d8c3d0042f720befd39f66ec9f747

brharrington:
On jdk9+ the String internals changed and it is more efficient to just use the String contructor that takes a byte array when dealing with ASCII values. Since jdk11 is the most recent LTS version with these changes and there are some methods added to String that can be used to detect it, this change uses the String constructor directly when on a newer JDK. The old behavior is preserved for jdk8.

Fixes #342

### DIFF
--- a/smile/src/main/java/com/fasterxml/jackson/dataformat/smile/SmileParser.java
+++ b/smile/src/main/java/com/fasterxml/jackson/dataformat/smile/SmileParser.java
@@ -3,6 +3,7 @@ package com.fasterxml.jackson.dataformat.smile;
 import java.io.*;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
 import com.fasterxml.jackson.core.*;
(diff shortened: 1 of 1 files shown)


brharrington:
From my tests it seems to maintain roughly the same performance on jdk8 and shows improvements for jdk11+.

```
Benchmark                       Mode  Cnt       Score       Error   Units

java 8 baseline                thrpt    5  348058.136 ± 29095.348   ops/s
java 8 with change             thrpt    5  342844.462 ± 32386.599   ops/s

java 11 baseline               thrpt    5  313255.865 ± 43280.554   ops/s
java 11 with change            thrpt    5  347304.547 ± 44594.978   ops/s

java 17 baseline               thrpt    5  322069.060 ± 28697.702   ops/s
java 17 with change            thrpt    5  427788.567 ± 25312.986   ops/s
```

cowtowncoder:
Looks good, and I think can be merged in 2.14 for 2.14.1 patch.

One thing I may need: if we don't yet have a CLA (I feel we might have but didn't see one?), it'd be from here:

and the usual way is to print it out, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once we have it (please LMK if you did send it earlier and I'll try to locate it) I can go ahead merge this nice improvement in.

cowtowncoder (approved):

PR #346, opened 2022-12-06
add tests for large nums

pjfanning:
These tests succeed despite the the large nums.
This is because the numbers are represented in binary form and the text length checks do not get invoked.
The question is whether we want to check the num lengths when we parse the binary nums.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -888,6 +888,7 @@ public class CBORParser extends ParserMinimalBase
         if (_binaryValue.length == 0) {
             _numberBigInt = BigInteger.ZERO;
         } else {
+            _ioContext.streamReadConstraints().validateIntegerLength(_binaryValue.length);
             BigInteger nr = new BigInteger(_binaryValue);
             if (neg) {
                 nr = nr.negate();
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -2140,8 +2140,10 @@ public class CBORParser extends ParserMinimalBase
         if ((_numTypesValid & (NR_DOUBLE | NR_FLOAT)) != 0) {
             // Let's parse from String representation, to avoid rounding errors that
             //non-decimal floating operations would incur
+            final String text = getText();
+            this.streamReadConstraints().validateFPLength(text.length());
             _numberBigDecimal = NumberInput.parseBigDecimal(
-                    getText(), isEnabled(StreamReadFeature.USE_FAST_BIG_NUMBER_PARSER));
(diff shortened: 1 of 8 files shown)


cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -2140,8 +2140,10 @@ protected void convertNumberToBigDecimal() throws IOException
         if ((_numTypesValid & (NR_DOUBLE | NR_FLOAT)) != 0) {
             // Let's parse from String representation, to avoid rounding errors that
             //non-decimal floating operations would incur
+            final String text = getText();
```
cowtowncoder, line 2143: Hmmmh. I am not sure the original logic makes much sense. We must have decoded underlying FP and seems like going into String-representation does nothing to improve or maintain accuracy... 

But... let me actually add comments at higher level as main thread, wrt limits

cowtowncoder:
Ok, so.... I think that.

1. We DO want to maintain number length limits, but
2. Limits should be imposed on binary encoded numbers (which will effectively allow bigger numbers since binary encoding is more efficient), NOT on String representation (unless there are specific cases where String representation must be caught)
3. We could consider different default length limits for text-vs-binary. But I think we can start by using the same, worrying later if need be.

Given this, I think length limits need to be caught at decoding time, not when converting. So tests are probably fine, but I think implementation needs to be different.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
pjfanning: this code already uses getText() - so I'm just applying the string length check here

There is also unmodified code for handling binary encoded numbers. I haven't changed that code yet and it would be good to do this too - but it isn't as obvious how to check the len in that flow.

An example from the Smile parsing code:
```
        if (_decode7BitEncoded()) { // got it all!
            // note: scale value is signed, needs zigzag, so:
            int scale = SmileUtil.zigzagDecode((int) _pending64);
            BigInteger bigInt = new BigInteger(_byteArrayBuilder.toByteArray());
            _numberBigDecimal = new BigDecimal(bigInt, scale);
            _numberType = NumberType.BIG_DECIMAL;
            _numTypesValid = NR_BIGDECIMAL;
            return _valueComplete(JsonToken.VALUE_NUMBER_FLOAT);
        }
```

So we could check the len of the _byteArrayBuilder.toByteArray() and fail if it is too long.

pjfanning:
@cowtowncoder I added some checks where the binary encoded numbers are decoded. I've updated the tests to match.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder: @pjfanning Right, yes, it is existing handling, did not think you were changing. So more a note to myself. :)

cowtowncoder (approved):

cowtowncoder:
Thank you @pjfanning !

PR #347, opened 2022-12-10
Add support for CBOR stringref extension.

here-abarany:
See http://cbor.schmorp.de/stringref for stringref definition. Repeated binary and text strings are assigned an index on write and a tagged integer is written when repeated. Parser respects the stringref IDs when reading data back in.

Nested tags are now supported. The previous getCurrentTag() function is still supported for backwards compatibility, and returns the first tag. Otherwise a new TagList type is used to efficiently handle multiple tags.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
@@ -1,9 +1,12 @@
 package com.fasterxml.jackson.dataformat.cbor;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.io.*;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.HashMap;
 
 import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.base.GeneratorBase;
@@ -78,13 +81,27 @@ public class CBORGenerator extends GeneratorBase
          * incoming String should fail with an exception or silently be output
          * as the Unicode 'REPLACEMENT CHARACTER' (U+FFFD) or not; if not,
          * an exception will be thrown to indicate invalid content.
-         *<p>
+         * <p>
          * Default value is {@code false} (for backwards compatibility) meaning that
          * an invalid surrogate will result in exception ({@link IllegalArgumentException}
          *
          * @since 2.12
          */
         LENIENT_UTF_ENCODING(false),
+
+        /**
+         * Feature that determines if string references are generated based on the
+         * <a href="http://cbor.schmorp.de/stringref">stringref</a>) extension. This can save
+         * storage space, parsing time, and pool string memory when parsing. Readers of the output
+         * must also support the stringref extension to properly decode the data. Extra overhead may
+         * be added to generation time and memory usage to compute the shared binary and text
+         * strings.
+         * <p>
+         * Default value is {@code false} meaning that the stringref extension will not be used.
+         *
+         * @since 2.15
+         */
+        STRINGREF(false),
         ;
 
         protected final boolean _defaultState;
@@ -175,7 +192,7 @@ public class CBORGenerator extends GeneratorBase
     /**********************************************************
      */
 
-	/**
+    /**
      * Intermediate buffer in which contents are buffered before being written
      * using {@link #_out}.
      */
@@ -236,6 +253,12 @@ public class CBORGenerator extends GeneratorBase
      */
     protected boolean _bufferRecyclable;
 
+    /**
+     * Table of previously referenced text and binary strings when the STRINGREF feature is used.
+     * @since 2.15
+     */
+    protected HashMap<Object, Integer> _stringRefs;
+
     /*
     /**********************************************************
     /* Life-cycle
@@ -255,6 +278,7 @@ public class CBORGenerator extends GeneratorBase
         _ioContext = ctxt;
         _out = out;
         _bufferRecyclable = true;
+        _stringRefs = Feature.STRINGREF.enabledIn(formatFeatures) ? new HashMap<>() : null;
         _outputBuffer = ctxt.allocWriteEncodingBuffer(BYTE_BUFFER_FOR_OUTPUT);
         _outputEnd = _outputBuffer.length;
         _charBuffer = ctxt.allocConcatBuffer();
@@ -292,6 +316,7 @@ public class CBORGenerator extends GeneratorBase
         _bufferRecyclable = bufferRecyclable;
         _outputTail = offset;
         _outputBuffer = outputBuffer;
+        _stringRefs = Feature.STRINGREF.enabledIn(formatFeatures) ? new HashMap<>() : null;
         _outputEnd = _outputBuffer.length;
         _charBuffer = ctxt.allocConcatBuffer();
         _charBufferLength = _charBuffer.length;
@@ -488,6 +513,17 @@ public class CBORGenerator extends GeneratorBase
         if (len == 0) {
             _writeByte(BYTE_EMPTY_STRING);
             return;
+        } else if (_stringRefs != null) {
+            // Check for a string reference.
+            String str = name.getValue();
+            Integer index = _stringRefs.get(str);
+            if (index != null) {
+                writeTag(TAG_ID_STRINGREF);
+                _writeIntMinimal(PREFIX_TYPE_INT_POS, index);
+                return;
+            } else if (shouldReferenceString(_stringRefs.size(), len)) {
+                _stringRefs.put(str, _stringRefs.size());
+            }
         }
         _writeLengthMarker(PREFIX_TYPE_TEXT, len);
         _writeBytes(raw, 0, len);
@@ -831,6 +867,17 @@ public class CBORGenerator extends GeneratorBase
         if (len == 0) {
             _writeByte(BYTE_EMPTY_STRING);
             return;
+        } else if (_stringRefs != null) {
+            // Check for a string reference.
+            String str = sstr.getValue();
+            Integer index = _stringRefs.get(str);
+            if (index != null) {
+                writeTag(TAG_ID_STRINGREF);
+                _writeIntMinimal(PREFIX_TYPE_INT_POS, index);
+                return;
+            } else if (shouldReferenceString(_stringRefs.size(), len)) {
+                _stringRefs.put(str, _stringRefs.size());
+            }
         }
         _writeLengthMarker(PREFIX_TYPE_TEXT, len);
         _writeBytes(raw, 0, len);
@@ -840,11 +887,24 @@ public class CBORGenerator extends GeneratorBase
     public void writeString(char[] text, int offset, int len)
             throws IOException {
         _verifyValueWrite("write String value");
+        String str = null;
         if (len == 0) {
             _writeByte(BYTE_EMPTY_STRING);
             return;
+        } else if (_stringRefs != null && len <= MAX_LONG_STRING_CHARS) {
+            // Check for a string reference.
+            str = new String(text, offset, len);
+            Integer index = _stringRefs.get(str);
+            if (index != null) {
+                writeTag(TAG_ID_STRINGREF);
+                _writeIntMinimal(PREFIX_TYPE_INT_POS, index);
+                return;
+            }
         }
-        _writeString(text, offset, len);
+        int actual = _writeString(text, offset, len);
+        if (str != null && shouldReferenceString(_stringRefs.size(), actual)) {
+            _stringRefs.put(str, _stringRefs.size());
+        }
     }
 
     @Override
@@ -855,6 +915,17 @@ public class CBORGenerator extends GeneratorBase
         if (len == 0) {
             _writeByte(BYTE_EMPTY_STRING);
             return;
+        } else if (_stringRefs != null) {
+            // Check for a string reference.
+            String str = new String(raw, offset, len, StandardCharsets.UTF_8);
+            Integer index = _stringRefs.get(str);
+            if (index != null) {
+                writeTag(TAG_ID_STRINGREF);
+                _writeIntMinimal(PREFIX_TYPE_INT_POS, index);
+                return;
+            } else if (shouldReferenceString(_stringRefs.size(), len)) {
+                _stringRefs.put(str, _stringRefs.size());
+            }
         }
         _writeLengthMarker(PREFIX_TYPE_TEXT, len);
         _writeBytes(raw, 0, len);
@@ -924,8 +995,25 @@ public class CBORGenerator extends GeneratorBase
             return;
         }
         _verifyValueWrite("write Binary value");
+        ByteBuffer bytesRef = null;
+        if (_stringRefs != null) {
+            bytesRef = ByteBuffer.wrap(data, offset, len);
+            Integer index = _stringRefs.get(bytesRef);
+            if (index != null) {
+                writeTag(TAG_ID_STRINGREF);
+                _writeIntMinimal(PREFIX_TYPE_INT_POS, index);
+                return;
+            }
+        }
+
         _writeLengthMarker(PREFIX_TYPE_BYTES, len);
         _writeBytes(data, offset, len);
+
+        if (bytesRef != null && shouldReferenceString(_stringRefs.size(), len)) {
+            // Store a copy of the data to ensure that modifications don't corrupt the lookup table.
+            _stringRefs.put(ByteBuffer.wrap(Arrays.copyOfRange(data, offset, len)),
+                    _stringRefs.size());
+        }
     }
 
     @Override
@@ -943,8 +1031,29 @@ public class CBORGenerator extends GeneratorBase
         _verifyValueWrite("write Binary value");
         int missing;
 
-        _writeLengthMarker(PREFIX_TYPE_BYTES, dataLength);
-        missing = _writeBytes(data, dataLength);
+        if (_stringRefs == null) {
+            _writeLengthMarker(PREFIX_TYPE_BYTES, dataLength);
+            missing = _writeBytes(data, dataLength);
+        } else {
+            // When computing string references must have the data available ahead of time.
+            byte[] bytes = new byte[dataLength];
+            missing = dataLength - data.read(bytes);
+            if (missing == 0) {
+                ByteBuffer bytesRef = ByteBuffer.wrap(bytes);
+                Integer index = _stringRefs.get(bytesRef);
+                if (index != null) {
+                    writeTag(TAG_ID_STRINGREF);
+                    _writeIntMinimal(PREFIX_TYPE_INT_POS, index);
+                } else {
+                    _writeLengthMarker(PREFIX_TYPE_BYTES, dataLength);
+                    _writeBytes(bytes, 0, dataLength);
+                    if (shouldReferenceString(_stringRefs.size(), dataLength)) {
+                        _stringRefs.put(bytesRef, _stringRefs.size());
+                    }
+                }
+            }
+        }
+
         if (missing > 0) {
             _reportError("Too few bytes available: missing " + missing
                     + " bytes (out of " + dataLength + ")");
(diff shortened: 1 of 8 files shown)


cowtowncoder:
Ok, first of all: thank you for contributing this feature! It sounds a useful thing, CBOR implementing support something Smile has had for a while.

One thing I was wondering about: since this is a sizable thing, I was wondering if it was possible to split this in two PRs: one dealing with parsing, and second one (building on first) for generation? I realize it may be tricky wrt tests, but it would be easier to code review that way if at all possible.

Second thing: regardless of whether it's 1 or 2 PRs, one (and only) process thing we need is CLA.
It can be found from here:

here-abarany:
> One thing I was wondering about: since this is a sizable thing, I was wondering if it was possible to split this in two PRs: one dealing with parsing, and second one (building on first) for generation?

From what I understand this would require separate branches, which would be very unwieldy to manage and would require the first PR to be submitted before the second could actually be looked at. (since the second one would contain the code from the first) What I can do is split it into two commits, which would allow you to review them separately while being part of the same branch/PR.

> Second thing: regardless of whether it's 1 or 2 PRs, one (and only) process thing we need is CLA.

Since I'm contributing this on company time I've forwarded this to the OSS department to make sure I get the OK to sign it. Hopefully it won't take too long to get approval.

cowtowncoder:
> > One thing I was wondering about: since this is a sizable thing, I was wondering if it was possible to split this in two PRs: one dealing with parsing, and second one (building on first) for generation?
> 
> From what I understand this would require separate branches, which would be very unwieldy to manage and would require the first PR to be submitted before the second could actually be looked at. (since the second one would contain the code from the first) What I can do is split it into two commits, which would allow you to review them separately while being part of the same branch/PR.
> 
> > Second thing: regardless of whether it's 1 or 2 PRs, one (and only) process thing we need is CLA.
> 
> Since I'm contributing this on company time I've forwarded this to the OSS department to make sure I get the OK to sign it. Hopefully it won't take too long to get approval.

Yes, would require 2 branches and I guess that won't play nicely with PRs. Too bad but I guess it makes sense to keep things bundled.

And yes, CLA should be fine; there's CCLA available if your company prefers that.

here-abarany:
Just to give a heads up, we have a tentative agreement internally to sign the CCLA, but are trying to find the right people to sign it. I'm hoping we'll get it submitted by mid January. In the meantime I'll be out the next two weeks myself for the holidays.

here-abarany:
@cowtowncoder I finally managed to get a signature and forwarded the CCLA last Friday. I just wanted to make sure that it was received and we can continue with the PR.

cowtowncoder:
@here-abarany I can't seem to find CCLA. Was it sent to `info` at fasterxml dot com?

here-abarany:
@cowtowncoder it was sent to `clas` rather than `info`, which was the address in the document. I can forward it to `info` as well.

cowtowncoder:
@here-abarany Ok odd. I think `clas` should get forward to `info`... but yes, please try re-sending it as I don't see it.

here-abarany:
@cowtowncoder I just forwarded it now.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -924,8 +958,25 @@ public void writeBinary(Base64Variant b64variant, byte[] data, int offset,
             return;
         }
         _verifyValueWrite("write Binary value");
+        ByteBuffer bytesRef = null;
+        if (_stringRefs != null) {
```
cowtowncoder, line 962: Does StringRefs functionality also apply to binary values? That seems.... odd. Not all binary values represent Strings.

Would it make sense to have separate features for enabling these separately? (I assume it is fine to write combination of references and non-referenced String / binary values).

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -245,6 +322,35 @@ private Feature(boolean defaultState) {
...
 
+    /**
+     * The depth of nested objects.
+     */
+    protected int _nestedDepth = 0;
+
+    /**
+     * Type to keep track of a list of string references. A depth is stored to know when to pop the
+     * references off the stack for nested namespaces.
+     */
```
cowtowncoder, line 333: Add `@since 2.15` for new classes, public methods in public API.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
here-abarany: The extension requires computing the index for every binary and text string within a namespace as long as it fits within the minimum size based on the current index value. (i.e. the more bytes required to store the index, the longer the threshold to compute the index for sharing) It doesn't allow you to selectively apply indices with the exception to indefinite length (chunked) strings, which will never have an index computed.

In other words, the only way to exclude binary strings from the index computation is to always write them with the indefinite length encoding.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -46,17 +48,92 @@ public static int collectDefaults()
...
+         * @return The first tag or -1 if there are no tags.
+         */
+        public int getFirstTag() {
+            if (_tagCount == 0) {
+                return -1;
+            }
+            return _tags[0];
+        }
+
+        private int[] _tags;
```
cowtowncoder, line 133: Maybe add `size()` method? (`return _tagCount;`)

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -501,42 +536,6 @@ public final void writeFieldId(long id) throws IOException {
...
-    /**********************************************************
-    /* Overridden methods, copying with tag-awareness
-    /**********************************************************
-     */
-
-    /**
-     * Specialize {@link JsonGenerator#copyCurrentEvent} to handle tags.
-     */
-    @Override
-    public void copyCurrentEvent(JsonParser p) throws IOException {
```
cowtowncoder, line 514: Where are these implemented now? Without overrides tag information will be lost if buffering (although TBH this won't work with general `TokenBuffer` anyway)

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -245,6 +322,35 @@ private Feature(boolean defaultState) {
...
+        }
+
+        public ArrayList<Object> stringRefs = new ArrayList<>();
+        public int depth;
+    }
+
+    /**
+     * Stack of text and binary string references.
+     */
+    protected Stack<StringRefList> _stringRefs = new Stack<>();
```
cowtowncoder, line 347: Could probably create sub-class that has helper methods, if that simplifies calling code (will add a note where this'd help)

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
here-abarany: This is explained in the message for the first commit in the series:

> Copying tags can be dangerous in two main cases:
> 1. Jackson creates its own tags, such as for BigNum or the self describe
   tag.
> 2. The semantic meaning of an element changes. For example, converting
   between stringrefs and no stringrefs.
>
> In some cases it would result in duplicate tags, while in others it may
cause completely incorrect tags, such as a stringref tag applying to a
string rather than an integer. Before nested tag support was added this
could generate files that Jackson could not parse, though even with nested
tag support it may cause undefined behavior.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -615,17 +732,25 @@ public JsonToken nextToken() throws IOException
...
                     _streamReadContext = _streamReadContext.getParent();
                     return (_currToken = JsonToken.END_OBJECT);
                 }
                 return (_currToken = _decodePropertyName());
             }
         } else {
             if (!_streamReadContext.expectMoreValues()) {
-                _tagValue = -1;
+                --_nestedDepth;
+                if (!_stringRefs.empty() && _stringRefs.peek().depth == _nestedDepth) {
```
cowtowncoder, line 750: Using a simple sub-class of `Stack` would allow hiding this logic into single method.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -837,6 +978,9 @@ protected String _numberToName(int ch, boolean neg) throws IOException
                 // [dataformats-binary#269] (and earlier [dataformats-binary#30]),
                 // got some edge case to consider
                 if (i < 0) {
+                    if (isStringref) {
+                        _reportError("String reference index too large.");
```
cowtowncoder, line 982: Please remove trailing `.`

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
cowtowncoder: Ok, not sure about part "Jackson creates its own tags" here (they come from CBOR spec)?

But aside from that, `copyCurrentEvent` is used to copy exact contents of input. I guess it can be problematic for shared String references as unlike other tags that only related to specific token/value without context/state, string-refs do have context (dictionary being collected).

Given that no unit test fails I guess this is fine. And if this becomes a problem can re-implement methods as necessary.

cowtowncoder:
Ok @here-abarany -- impressive changes! I had some questions etc, although nothing major.

My main concern is actually getting this to merge against 3.0 (`master`). I know it's not practical to split this into smaller chunks, but one thing that might help is if white-space changes (removal of white-space from end-of-lines) could be done as a separate PR first?
I could merge that in, then into `master`; that'd probably remove noise slightly.

I suspect I will need help with `master` merge anyway but maybe little bit less.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
here-abarany: I mean standard CBOR tags that are created based on the generation code within the `CBORGenerator` class.

This behavior is because the event forwarding still goes through the standard generator interface. For example, let's say you are reading a value that gets interpreted as a BigNum. A CBOR tag is created to decorate the binary string value that holds the data to denote that it represents a BigNum. This logic will be executed regardless of whether an existing BigNum tag exists or not, resulting in *two* BigNum tags. (one from the tag copy, one created from `CBORGenerator`)

As more tags are supported by Jackson, the more dangerous that behavior can get. For example, let's say you're converting from a stringref encoded CBOR to one without stringrefs. It would read an integer index corresponding to a string and convert it to a "real" string, but the tag copying logic it would *also* apply the tag denoting that the value is a stringref index. This is invalid as the stringref tag is intended to apply to an integer, while it would apply to a string instead.

here-abarany:
@cowtowncoder thanks for the review, I hope to push up an update later today based on your comments. I'll have two separate pushes: one to rebase onto the latest 2.15 changes and one to push up the updates.

With regards to the whitespace changes, my editor is configured to clear trailing whitespace on save. I might be able to check out to the 2.15 head, save those files, and cherry-pick my changes on top of that to reduce some of the noise. If I'm successful I'll include that with my pushes later today.

I can try merging into the master branch locally and see what happens. Hopefully everything merges cleanly, if not I'll note what issues I encountered.

cowtowncoder:
> @cowtowncoder thanks for the review, I hope to push up an update later today based on your comments. I'll have two separate pushes: one to rebase onto the latest 2.15 changes and one to push up the updates.
> 
> With regards to the whitespace changes, my editor is configured to clear trailing whitespace on save. I might be able to check out to the 2.15 head, save those files, and cherry-pick my changes on top of that to reduce some of the noise. If I'm successful I'll include that with my pushes later today.
> 
> I can try merging into the master branch locally and see what happens. Hopefully everything merges cleanly, if not I'll note what issues I encountered.

It's unlikely to merge cleanly, partly due to package renaming, partly due to big API changes.
But exactly how tricky it is can vary. :)

As to whitespace, I'll try to figure out how to do that on my end for 2.15; that way you can just rebase on 2.15 head. No need for you to have to do that now I think.

cowtowncoder:
@here-abarany Ok I did white-space trimming, merged it to `master` (re-linking some sources that hopefully reduces conflicts -- changing Java package causes lots of pain). So you may want to merge/rebase from 2.15 now.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
here-abarany: I did a quick check and added it to this and the new tag constants in CBORConstants. Let me know if I missed any others.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
here-abarany: I've added the method as requested.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
here-abarany: I created a helper class for this. I chose to use containment rather than subclass for this case as the interface is slightly different.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
here-abarany: I think I got all the added errors, let me know if I missed any.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
cowtowncoder: Ok. So it is required by CBOR spec. Interesting choices; esp. not having upper limit. Could be problematic for many binary data cases but I guess that's not my problem :)
(I spec'd/implemented String back refs for Smile format before CBOR was a thing; never even considered shared refs for binary data so this seems.... interesting).

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder: Thanks. I think it's fine; if some are missing I'll add a note.

here-abarany:
@cowtowncoder I have pushed up my changes, I believe I addressed all the comment threads from the review. The first push is just a rebase, so you should be able to view just my latest changes by comparing the latest push just before this comment.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
cowtowncoder: Ah. Ok, yes.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder: @here-abarany Ok yes, that is cleaner approach from OO design perspective. Although being only used internally not a big thing either way.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
here-abarany: I would imagine that if a client has a significant amount of binary data to the point where this becomes an issue they could simply not opt into stringrefs. If clients are concerned about *both* a significant number of shared strings bloating their data *and* large embedded binaries (or large strings for that matter), an option would be to introduce a configuration to have an upper limit on string size before it forces a chunked encoding. (which would bypass the stringref sharing) I would imagine this is very likely very rare, however, and probably not worth the complexity of introducing an option until there's a real-world example that both proves the need for it and provides an example to more effectively create such a feature.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -236,6 +253,11 @@ public int getMask() {
      */
     protected boolean _bufferRecyclable;
 
+    /**
+     * Table of previously referenced text and binary strings.
+     */
```
cowtowncoder, line 258: ... Used if StringRef functionality is enabled. // or something like that

`@since 2.15`

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -245,6 +331,64 @@ private Feature(boolean defaultState) {
...
+    }
+
+    /**
+     * Stack of text and binary string references.
+     */
+    protected StringRefListStack _stringRefs = new StringRefListStack();
+
+    /**
+     * Shared string that should be used in place of _textBuffer when a string reference is used.
+     */
```
cowtowncoder, line 389: @since 2.15

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
here-abarany: I wasn't sure if protected fields should be marked, I'll go through and try to add it for the fields I've added. (as well as the comment clarification noted here)

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/ParserWithJsonOrgSampleTest.java
```diff
@@ -47,6 +47,16 @@ public void testJsonSampleDoc() throws IOException
         verifyJsonSpecSampleDoc(cborParser(data), false, true);
     }
 
+    public void testJsonSampleDocStringref() throws IOException
+    {
+        byte[] data = cborDoc(new CBORFactory().enable(CBORGenerator.Feature.STRINGREF),
```
cowtowncoder, line 52: Could you change this to use Builder? This merges cleanly with 3.0 where factories are immutable.

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/mapper/BiggerDataTest.java
```diff
@@ -137,4 +140,33 @@ public void testRoundTrip() throws Exception
 		assertEquals(citm.topicSubTopics.size(), citm2.topicSubTopics.size());
 		assertEquals(citm.venueNames.size(), citm2.venueNames.size());
 	}
+
+	public void testRoundTripStringref() throws Exception
+	{
+		Citm citm0 = MAPPER.readValue(getClass().getResourceAsStream("/data/citm_catalog.json"),
+				Citm.class);
+		ObjectMapper mapper = new CBORMapper(
+		    new CBORFactory().enable(CBORGenerator.Feature.STRINGREF));
```
cowtowncoder, line 149: Same here, builder over factory configuration (I know, many existing tests use this, trying to move over with new code).

cowtowncoder:
@here-abarany Ok, sorry, missed couple of smaller things; now done with reviewing I think. Minor test changes requested to make it more likely things merge less uncleanly.

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/ParserWithJsonOrgSampleTest.java
here-abarany: I've switched this and the one below to use a builder.

here-abarany:
@cowtowncoder I have pushed up a new revision to fix your latest comments.

cowtowncoder:
Thank you @here-abarany! I will pick this up again today or tomorrow, get it merged.

cowtowncoder (approved):

cowtowncoder:
Ok, phew. Merge to `master` went ok, almost: there is just one new failure and that's for String-Refs; `BiggerDataTest.testRoundTripStringref()`.

I suspect it is since `CBORParser.nextFieldName()` methods (renamed as just `nextName()` in 3.0) do not all handle String-refs.
Or rather, new `nextNameMatch()` method doesn't -- it probably needs to, and I think `jackson-databind` uses that for Bean properties in 3.0.

Could you have a look if you can see how to fix this, @here-abarany ?

here-abarany:
@cowtowncoder thanks for getting it merged in, I'll see if I can take a look on Monday for the master issue.

cowtowncoder:
> @cowtowncoder thanks for getting it merged in, I'll see if I can take a look on Monday for the master issue.

Thank you in advance! Hopefully it is relatively straight-forward.

PR #355, opened 2023-02-07
Fix handling of stringrefs for CBORParser.nextNameMatch()

here-abarany:
Read tags to properly identify stringrefs and take into account strings that need to be added to the stringref table.


here-abarany:
@cowtowncoder this fixes the CBOR tests on the master branch. It's good that this issue was caught based on the test coverage, it might be a good idea to double check whether any other code paths might have been missed with the test coverage currently in place.

cowtowncoder (approved):

cowtowncoder:
Thank you @here-abarany ! Yes, it is good there was at least one failing test. It'd be good to retrofit other existing tests to have alternative of "standard" handling where string-refs are enabled for round-trip handling

PR #356, opened 2023-02-07
Add WRITE_MINIMAL_DOUBLES feature to CBORGenerator

here-abarany:
Enabling WRITE_MINIMAL_DOUBLES works similarly to WRITE_MINIMAL_INTS, choosing to use a smaller data representation (single-precision float) when the conversion will result in no loss of data. This can provide a substantial savings for data that contains many doubles that don't require full precision.

Cleaned up some of the writing code for floats and doubles to avoid code duplication, and reserve the proper amount of space for output for both types.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
@@ -102,6 +102,19 @@ public class CBORGenerator extends GeneratorBase
          * @since 2.15
          */
         STRINGREF(false),
+
+        /**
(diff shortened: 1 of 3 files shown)


here-abarany:
@cowtowncoder Just wanted to make sure you saw this PR. It's a relatively small addition, especially compared to my previous submission.

cowtowncoder (approved):

cowtowncoder:
Excellent, @here-abarany -- sounds like useful feature, simple enough implementation. Could even consider possibility of porting to Smile format backend. But for now I'll merge it.

cowtowncoder:
Ok, merged in 2.15 then to master.

here-abarany:
Thanks!

PR #363, opened 2023-03-08
fix for oss fuzz issue 35979

pjfanning:
cbor issue - https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=35979


cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -2549,68 +2549,72 @@ private final String _finishShortText(int len) throws IOException
...
-        String str = _textBuffer.setCurrentAndReturn(outPtr);
-        if (stringRefs != null) {
-            stringRefs.stringRefs.add(str);
-            _sharedString = str;
+            return str;
+        } catch (RuntimeException re) {
+            throw new IOException("Failed to parse CBOR input", re);
         }
-        return str;
     }
```
pjfanning, line 2618: despite the Github diff - this is just the same code but wrapped in a try-catch

cowtowncoder: Yeah. But I don't think defensive programming makes sense here. I think instead code in question has to check for buffer boundaries and not count on JVM's array checks; the big problem being that there is no information on what in content is broken. (if I remember correctly, having looked that, this is buffer overrun situation; split UTF-8 codepoint or such).

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/fuzz/Fuzz35979ShortTextTest.java
```diff
@@ -0,0 +1,32 @@
+package com.fasterxml.jackson.dataformat.cbor.fuzz;
```
cowtowncoder, line 1: There is already 

com/fasterxml/jackson/dataformat/cbor/failing/Fuzz_35979_StringValueTest.java

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
pjfanning: * I guess that it is optimal to go and check the bounds instead of catching the exception.
* One advantage to catching the exception is that you don't make the happy path slower by adding additional range checks.
* It does get rid of the annoying public oss-fuzz issue and could be treated as an interim solution until someone gets into the nitty gritty of the code and finds a performant way of checking the array bounds.

cowtowncoder: My concern is mostly that the specific bug could (if I recall correctly) cause corrupt data and NOT hit ArrayIndexOutOfBoundsException, depending on how input is given.

I agree oss-fuzz issue would be nice to resolve, one of the last ones. I fixed a few of these a year or more ago but not this one.

PR #364, opened 2023-03-08
avro nesting checks

pjfanning:
* I'm not sure if Avro can be exploited by a nested data attack.
* This diff is fairly large
* I still need to work out how to test this

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/RecordReader.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/RecordReader.java
@@ -22,15 +22,18 @@ abstract class RecordReader extends AvroStructureReader
     protected final int _count;
 
     protected RecordReader(AvroReadContext parent, AvroFieldReader[] fieldReaders, AvroParserImpl parser, String typeId)
+            throws IOException
     {
         super(parent, TYPE_OBJECT, typeId);
         _fieldReaders = fieldReaders;
         _parser = parser;
         _count = fieldReaders.length;
+        if (parser != null)
+            parser.streamReadConstraints().validateNestingDepth(_nestingDepth);
     }
 
     @Override
-    public abstract RecordReader newReader(AvroReadContext parent, AvroParserImpl parser);
+    public abstract RecordReader newReader(AvroReadContext parent, AvroParserImpl parser) throws IOException;
 
     @Override
     public String getCurrentName() { return _currentName; }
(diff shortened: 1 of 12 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/RecordReader.java
```diff
@@ -22,15 +22,18 @@ abstract class RecordReader extends AvroStructureReader
     protected final int _count;
 
     protected RecordReader(AvroReadContext parent, AvroFieldReader[] fieldReaders, AvroParserImpl parser, String typeId)
+            throws IOException
     {
         super(parent, TYPE_OBJECT, typeId);
         _fieldReaders = fieldReaders;
         _parser = parser;
         _count = fieldReaders.length;
+        if (parser != null)
```
cowtowncoder, line 31: Can that be `null`? Maybe for some tests?

cowtowncoder:
Hmmh. Whether Avro Schemas allow unlimited nesting... I am not 100% sure.

cowtowncoder:
I think it's a good idea to see if a test can be created; and if not maybe postpone until someone actually reports the problem.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/RecordReader.java
pjfanning: a lot of tests seem to fail without this

pjfanning:
https://www.cvedetails.com/vulnerability-list/vendor_id-45/product_id-76652/year-2022/Apache-Avro.html -- seems that Avro's own impls have had CVEs reported. May not be related to problematic inputs.

cowtowncoder:
@pjfanning Too bad descriptions do not really show much about actual issue, wrt implementation. Seems more likely some validation of things like encoded lengths is not done.

cowtowncoder:
Ok, I think Avro does allow recursive types; f.ex this

https://kb.databricks.com/en_US/data-sources/recursive-references-in-avro-schema-are-not-allowed

suggest that simple self-references are allowed. I think same goes for Protobuf as well.
But being schema-based, I think that potential for DoS is lower because attacker must target specific schema type.
Still, I assume tree-style structures are fairly common so maybe it makes sense to add checks, given that it is not a ton of code (more so for Avro just to expose IOException but otherwise).

pjfanning:
@cowtowncoder I'm happy to merge this as is but I'm not sure I have the Avro expertise to produce a test case. I think you are right that having sensible AvroSchemas should protect users anyway.

The gain from adding this PR is that the exceptions are exposed. The implementation can be changed if anyone can provide a real use case that the code doesn't protect against. The values are configurable, so if someone gets a StreamConstaintsException and they still think the input is valid - they can increase the limits by tuning the StreamReadConstraints.

The potential loss is that this PR could turn out to have bugs in it that mean StreamReadConstraints tuning won't help.

cowtowncoder (approved):

PR #365, opened 2023-03-08
Delete avro-1.9.patch

pjfanning:
I can't work out if this is needed - but it looks like it is an old patch file that was committed to git by mistake.


cowtowncoder (approved):
Almost certainly accident. Thanks!

PR #367, opened 2023-03-11
Add unit test for too deep nesting check for Avro read

cowtowncoder:


cowtowncoder:
@pjfanning Adding test for nesting check wrt Avro -- had to use lower nesting since Avro generator can't write 1500 levels due to StackOverflow :)

I think similar thing would work for Protobuf too.

pjfanning:
@cowtowncoder should we lower the default limit on StreamReadConstraints - to 200, maybe?

Otherwise, we'll need to document that Avro users should explicitly set the 200 limit themselves.

cowtowncoder:
Not on account of my test: the problem was on generation side, not decoding.
(although maybe worth considering for generation-side limits once we get there...)

But if we could establish that Avro decoding has lower maximum then we probably should consider that.
Until then I think it's useful to keep maximum settings the same across formats.

PR #370, opened 2023-03-14
remove unused method

pjfanning:
I know this method is public but its name starts with an underscore and there is no javadoc - very unlikely any external users use it and the internal code doesn't use it.


cowtowncoder (approved):

cowtowncoder:
Usually would deprecate etc, but I concur with @pjfanning's assessment.

PR #371, opened 2023-04-06
validate big decimal scale before converting to big-int

pjfanning:


cowtowncoder (approved):

cowtowncoder:
I wonder if I should demote `_convertBigDecimalToBigInteger()` from `ParserBase` to `ParserMinimalBase`...

PR #386, opened 2023-07-08
Add support for `StreamWriteConstraints` checks

pjfanning:
relates to https://github.com/FasterXML/jackson-core/pull/1055


cowtowncoder (approved):

cowtowncoder:
Phew. Managed to merge to `master`.

One quick note: limits will not work with Avro serialization as of now -- problem is that we construct Avro's internal objects using separate stack. If anyone has time and interest, this is solvable; just need to add checks in bit different places.

I'll actually create a new issue for this flaw.

PR #391, opened 2023-09-08
Support enum avro serialization with default value

### LINKED ISSUE #388: JsonEnumDefaultValue not supported when using AvroMapper to generate schema from JAVA class
```
public enum Foobar {
    @JsonEnumDefaultValue
    FOO,
    BAR;
}

produces 
{
  "type" : "enum",
  "name" : "Foobar",
  "namespace" : "foo",
  "symbols" : [ "FOO", "BAR" ]
}
```

Need it to have the default key value pair in order to make enum forward compatible

```
{
  "type" : "enum",
  "name" : "FooBar",
  "namespace" : "foo",
  "symbols" : [ "FOO", "BAR" ],
  "default" : "FOO"
}
```

Unless I'm missing something I can't find a way to configure the class to fill in the default value when creating the schema

Configured the AvroMapper  in some groovy to create the schema for reference

```
      avroMapper.configure(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES, false)
                    .configure(JsonGenerator.Feature.IGNORE_UNKNOWN, true);
          avroMapper.configure(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE, true)

      AvroSchemaGenerator gen = new AvroSchemaGenerator();
        gen.enableLogicalTypes();
        avroMapper.acceptJsonFormatVisitor(classFile, gen);
        AvroSchema schemaWrapper = gen.getGeneratedSchema();

        org.apache.avro.Schema avroSchema = schemaWrapper.getAvroSchema();
        String avroSchemaInJSON = avroSchema.toString(true);
```

Sonic-Rage:
Fixes #388.


avro/pom.xml
```diff
@@ -47,7 +47,7 @@ abstractions.
     <dependency>
       <groupId>org.apache.avro</groupId>
       <artifactId>avro</artifactId>
-      <version>1.8.2</version>
+      <version>1.9.2</version>
```
cowtowncoder, line 50: I was under impression that this will break about everything... did you run unit test suite?

But even if not, we need this to be in separate PR, with separate issue.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
```diff
@@ -71,6 +72,14 @@ public abstract class AvroSchemaHelper
             String.class
     ));
 
+    /**
+     * 
+     * Jackson annotation introspector for verifying enum default annotation for Avro Schema generation
+     * 
+     * @since 2.16
+     */
+    private static final JacksonAnnotationIntrospector JACKSON_ANNOTATION_INTROSPECTOR = new JacksonAnnotationIntrospector();
```
cowtowncoder, line 81: No, `AnnotationIntrospector` should be passed from outside; it may be reconfigured by caller.
We should not rely on default instance.

cowtowncoder:
Ok, I realized that the avro core dependency upgrade to 1.9.x would be needed to access "enum default" information (both create and get). But this upgrade is not possible since `jackson-dataformat-avro` code itself is incompatible with 1.9.x (see #167 f.ex).

So I am not 100% sure this is actually achievable at this point.

I did make couple of changes to PR, but it is unfortunately not in a compilable state.

avro/pom.xml
Sonic-Rage: I was wary of that upgrade as well, but doing that version bump and running the avro test suite resulted in passing tests. 

We're a gradle shop so running stuff with maven was a bit hokey with all our firewall rules.   Readying through #167 looks like some stuff has been upgraded but I imagine there must be missing unit tests for some of the underlying avro lib changes.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaHelper.java
Sonic-Rage: makes sense.  thanks

avro/pom.xml
cowtowncoder: Ok, that's odd wrt tests: for me code did not compile with missing imports.
And compilation on CI and locally did not work.

I wish I was more up to date with Avro lib changes since there are newer minor versions, some of which might resolve issues -- although from what I remember, things were quite messy wrt backwards-compatibility (lack thereof) for some specific usage.

cowtowncoder:
Ok, changed target to 2.18, merged changes and updated call of deprecated method to non-deprecated.

@Sonic-Rage This might be ready for merge, but I think you should have a look.

One thing before merging tho: I think I'd need a CLA (https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf) if you didn't send one earlier.
It's usually easiest done by printing the doc, fill & sign, scan/photo, email to `cla` at fasterxml dot com. Only needs to be done once, before the first contribution merged; one is good for all future contributions.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaGenerationTest.java
```diff
@@ -169,4 +172,38 @@ public void testSchemaForUntypedMap() throws Exception
             verifyException(e, "Maps with non-stringable keys are not supported (yet?)");
         }
     }
+
+    // Issue 388 Default value for enums with class
+    public void testClassEnumWithDefault() throws Exception
```
MichalFoksa, line 177: I think this test better belongs to `EnumTest` class.

`EnumTest` did not exists at the time of this PR yet.

cowtowncoder: Will move.

cowtowncoder:
@MichalFoksa Looks like @Sonic-Rage has not been active in quite a while. PR looks legit to me; do you think you could sanity check it too? If so, I could merge it in 2.19 for 2.19.0.

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/EnumTest.java
```diff
@@ -127,4 +150,34 @@ public void test_avroSchemaWithString_fromStringValueToEnumValue() throws Except
         assertEquals(Gender.F, output.gender);
     }
 
+    // [dataformats-binary#388]: Default value for enums with class
+    public void testClassEnumWithDefault() throws Exception
```
MichalFoksa, line 154: @cowtowncoder 
Test is missing `@Test` annotation.

MichalFoksa (commented):
Just that missing `@Test` annotation, otherwise PR is fine

avro/src/test/java/com/fasterxml/jackson/dataformat/avro/EnumTest.java
cowtowncoder: Good catch, will add!

cowtowncoder:
@MichalFoksa Thank you!

MichalFoksa:
You are welcome!

PR #402, opened 2023-10-03
Fix #399; support SmileBufferRecycler via new 2.16 pluggable mechanism

### LINKED ISSUE #399: Rewrite Smile buffer recycling (`SmileBufferRecycler`) to use new (2.16) `RecyclerPool`, allow configuring use of non-ThreadLocal based pools
Now that `jackson-core` added pluggability of recycler pools (see https://github.com/FasterXML/jackson-core/issues/1089 ), and also provides a set of reusable implementations (see https://github.com/FasterXML/jackson-core/pull/1064) we can tackle Smile-specific additional recycling: it also needs to allow re-configuring esp. to work with Project Loom where `ThreadLocal` based pooling really does not work well. This is likely to also help with Android usage, where `SoftRefernence`s (used by default Smile recycler, in addition to `ThreadLocal`) does not work as well as vanilla J2SE JDKs.

So let's try to get this done for 2.16.

Class to change is `SmileBufferRecycler`.
For 2.16 we should still default to traditional recycler, but for 2.17+ should probably change the default. This allows for early adopters to switch over sooner, while we get feedback on how pooling works, best default to use.

cowtowncoder:
As per title, if we want to continue buffer recycling, here are changes for minimal upgrade.

But based on local performance testing it is not certain recycling is worth it -- will next consider that.

### DIFF
--- a/smile/src/main/java/com/fasterxml/jackson/dataformat/smile/SmileBufferRecycler.java
+++ b/smile/src/main/java/com/fasterxml/jackson/dataformat/smile/SmileBufferRecycler.java
@@ -1,7 +1,10 @@
 package com.fasterxml.jackson.dataformat.smile;
 
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.fasterxml.jackson.core.util.RecyclerPool;
+import com.fasterxml.jackson.core.util.RecyclerPool.WithPool;
 import com.fasterxml.jackson.dataformat.smile.SmileGenerator.SharedStringNode;
 
 /**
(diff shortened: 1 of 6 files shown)


pjfanning:
@cowtowncoder is this going to dropped out of 2.16 release? The RC1 has been published and ideally we wouldn't want to sneak in extra changes unless necessary.

cowtowncoder:
@pjfanning I decided to instead drop buffer recycling for Smile altogether, so this PR is not necessary -- but I kept it around as sample in case similar was needed for say Avro. But I can now close this as "won't do". Thanks for the reminder.

PR #404, opened 2023-10-03
Fix #403: remove SmileBufferRecycler from 2.16

### LINKED ISSUE #403: Remove Smile-specific buffer-recycling
(note: counter-proposal to #399 ) 

Some basic performance testing suggests that existing Smile-specific buffer recycling might be of limited value, especially for default settings (with only name back references).

So let's do a PR to see how performance actually behaves.

cowtowncoder:
So, alternate take to #399 where we just remove `SmileBufferRecycler` since it appears not to bring significant performance boost.


cowtowncoder:
Ok I think that for simplicity's sake, this makes more sense. Can reconsider in future if need be.

PR #408, opened 2023-10-25
Clean up of Ion module

thetric:


avro/pom.xml
```diff
@@ -54,7 +54,7 @@ abstractions.
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
-      <version>1.2.11</version>
```
cowtowncoder, line 57: Ideally shouldn't add unrelated changes, but since it's test dependency, this one is fine.

cowtowncoder:
Thank you @thetric !

Ok, nice minor clean ups which could go in 2.15. But I think the version upgrade needs to go in 2.16, just in case -- I don't expect too many issues with Ion upgrade but just to be safe.

What I can do is take this PR, make other changes in 2.15 and create separate PR for Ion dep version change.

cowtowncoder (approved):

PR #418, opened 2023-12-05
Wrap possible AssertionError from Ion class

### LINKED ISSUE #417: IonReader classes contain assert statement which could throw unexpected AssertionError
In the `IonParser::getText()` method, there is a call to the `IonReader::stringValue()` which is served by an Amazon implementation of `IonReaderTextSystemX`. The method does throw `UnknownSymbolException` if the symbol id cannot be resolved. But it also contains some assert statements which throw `AssertionError` when the resolved symbol id is 0 or negative. The AssertionError`` is not handled and is thrown to the users unexpectedly.
```java
    @Override
    public String getText() throws IOException
    {
         if (_currToken != null) { // null only before/after document
          ......
            case VALUE_STRING:
                try {
                    // stringValue() will throw an UnknownSymbolException if we're
                    // trying to get the text for a symbol id that cannot be resolved.
                    return _reader.stringValue();
                } catch (UnknownSymbolException e) {
                    throw _constructError(e.getMessage(), e);
                }
            ......           
        return null;
    }
```

The simplest fix is to also catch the `AssertionError`, the same as the `UnkonwnSymbolException`. In general, AssertionError should be internal use only and should be wrapped and avoided by throwing directly to the users. Not sure if it is meant to not handle it in this situation.

We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64721.

arthurscchan:
This PR fixes #417 by wrapping the unexpected AssertionError with the general JsonParseException to avoid throwing AssertionError directly to the users when invalid input is provided.

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -274,8 +274,11 @@ public class IonParser
                 try {
                     // stringValue() will throw an UnknownSymbolException if we're
                     // trying to get the text for a symbol id that cannot be resolved.
+                    // stringValue() has an assert statement which could throw an
+                    // AssertionError if we're trying to get the text with a symbol
+                    // id less than or equals to 0.
                     return _reader.stringValue();
-                } catch (UnknownSymbolException e) {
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):
LGTM

PR #419, opened 2023-12-06
build(ion): update to Amazon Ion 1.11.0

thetric:


cowtowncoder:
Sorry missed this PR; updated concurrently so 2.17 has the new version. Seems to have fixed some OSS-Fuzz reported issues.

Issue #424, opened 2023-12-18
`IonReader` throws `NullPointerException` for unchecked invalid data


arthurscchan:
In the `IonParser::getText()` method, there is a call to the `IonReader::stringValue()`. Also, in `IonParser::getXXXValue()` for retrieving different number values from the `IonReader` calls to underlying `IonReader` for retrieving string or number value. According to the Javadoc of `IonReader`, each of the APIs requires a special `IonType` and `IllegalStateException` could be thrown if the wrong type is passed. But there is a special case when there is no more input, the `IonType` will be null and continuing calling those methods will result in `NullPointerException`.  

```java
    @Override
    public String getText() throws IOException
    {
         if (_currToken != null) { // null only before/after document
          ......
            case VALUE_STRING:
                try {
                    // stringValue() will throw an UnknownSymbolException if we're
                    // trying to get the text for a symbol ID that cannot be resolved.
                    return _reader.stringValue();
                } catch (UnknownSymbolException e) {
                    throw _constructError(e.getMessage(), e);
                }
            ......           
        return null;
    }
...
    @Override
    public BigInteger getBigIntegerValue() throws IOException {
        return _reader.bigIntegerValue();
    }

    @Override
    public BigDecimal getDecimalValue() throws IOException {
        return _reader.bigDecimalValue();
    }

    @Override
    public double getDoubleValue() throws IOException {
        return _reader.doubleValue();
    }

    @Override
    public float getFloatValue() throws IOException {
        return (float) _reader.doubleValue();
    }

    @Override
    public int getIntValue() throws IOException {
        return _reader.intValue();
    }

    @Override
    public long getLongValue() throws IOException {
        return _reader.longValue();
    }
```

It is found that in the `IonParser::getNumberValue()` method, there is a null check to ensure the `IonType` (and `NumberType`) of the current token is not null before calling the corresponding data retrieving method in the `IonReader` implementation. But these null checks are missing from the above method which could cause unexpected `NullPointerException`. 

```java
    @Override
    public Number getNumberValue() throws IOException {
        NumberType nt = getNumberType();
        if (nt != null) {
            switch (nt) {
            case INT:
                return _reader.intValue();
            case LONG:
                return _reader.longValue();
            case FLOAT:
                return (float) _reader.doubleValue();
            case DOUBLE:
                return _reader.doubleValue();
            case BIG_DECIMAL:
                return _reader.bigDecimalValue();
            case BIG_INTEGER:
[...]

PR #425, opened 2023-12-18
Fixes for #424: Add null checking and wrap NPE

arthurscchan:
This PR provides a fixes for #424 by adding some null checking and wrap possible NPE in `IonParser`.

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -332,56 +332,48 @@ public class IonParser
 
     @Override
     public BigInteger getBigIntegerValue() throws IOException {
-        NumberType nt = getNumberType();
-        if (nt == null) {
-            throw _constructError("Wrong number type: BigInteger not found.", null);
-        }
+        _verifyIsNumberToken();
         return _reader.bigIntegerValue();
     }
 
     @Override
     public BigDecimal getDecimalValue() throws IOException {
-        NumberType nt = getNumberType();
-        if (nt == null) {
-            throw _constructError("Wrong number type: BigDecimal not found.", null);
-        }
+        _verifyIsNumberToken();
         return _reader.bigDecimalValue();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -331,31 +338,55 @@ public int getTextOffset() throws IOException {
 
     @Override
     public BigInteger getBigIntegerValue() throws IOException {
+        NumberType nt = getNumberType();
```
cowtowncoder, line 341: I think I'd like handling of this part to be bit different: could you have a separate PR for `getXXXValue()` changes, so I can fix other NPE parts from above?
(I can suggest alternative way on that PR without slowing merging of this one)

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -286,6 +286,13 @@ public String getText() throws IOException
                         msg = "UNKNOWN ROOT CAUSE";
                     }
                     throw _constructError("Internal `IonReader` error: "+msg, e);
+                } catch (NullPointerException e) {
```
cowtowncoder, line 289: I think this could then be combined with above as

    } catch (AssertionError | NullPointerException e) {

(or whatever the syntax was to allow multiple exception types)
since handling is basically the same, for now

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
cowtowncoder: Actually never mind; simple enough to isolate check, updated PR and will merge. Thanks again!

arthurscchan: Thanks.

cowtowncoder: Thank you @arthurscchan for figuring out the root cause & providing fix!

Issue #426, opened 2023-12-18
`SmileParser` throws unexpected IOOBE for corrupt content


arthurscchan:
In the `SmileParser::nextTextValue()` method, there is a line that uses the Integer `ptr` as an index to retrieve a byte from the `_inputBuffer`. But it is found that with some invalid input and repeat calling to the `SmileParser::nextTextValue()` method, it could cause ptr to be negative and trigger an unexpected `ArrayIndexOutOfBoundsException`.

```java
     public String nextTextValue() throws IOException
    {
       ...
            int ptr = _inputPtr;
            if (ptr >= _inputEnd) {
               ...
            }
            _tokenOffsetForTotal = ptr;
            int ch = _inputBuffer[ptr++] & 0xFF;
       ...
```
The simplest fix is to add a bound check for the ptr before using it as the array index.

We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65126.

cowtowncoder:
The problem here is that this might not be directly due to anything within `nextTextValue()` itself but (possibly) in preceding steps. So it is necessary to see the sequence of things that lead to the problematic state, in which call to `nextTextValue()` (and likely other calls) would fail.

Unfortunately I don't think `nextTextValue()` can truly validate offset at that point, but rather whatever lead to invalid value needs to be fixed
(specifically: just because offset is within valid buffer does not mean it might not be corrupt -- it being off the buffer does indicate it is invalid, of course, but the goal is prevent the problem where it occurs).

It is very likely that this requires an invalid document being read; but it may also rely on specific accessors/iteration methods being called.

PR #427, opened 2023-12-18
Fixes for #426: Add bound check for array index

arthurscchan:
This PR provides a suggested fix for #426 by adding a bound check before the `ptr` Integer used as the index for accessing the `_inputBuffer` array.


cowtowncoder:
Ok, so I think test is useful but I don't like the fix here -- it works around the problem but doesn't actually solve the problem.

Would it be possible to just add failing test (under `src/test/java/..../failing`) int this PR and I can see what goes wrong with decoding wrt sequence of calls.

cowtowncoder:
Interesting. The token in the beginning if `JsonToken.VALUE_EMBEDDED_OBJECT` representing binary data.
It's invalid, specifies huge length; but skipping produces invalid state.

cowtowncoder (approved):

arthurscchan:
@cowtowncoder Thanks for merging them in. I just had the chance to see your comments and you already fixed that for me. Thanks for your prompt reply. Seems that it justifies the use of fuzzers, especially for these large, strange and malformed inputs.

cowtowncoder:
Yes, I think fuzzing is good as long as we can replicate their findings. Picks up many edge cases.

Issue #432, opened 2023-12-27
More methods from `IonReader` could throw an unexpected `AssertionError`


arthurscchan:
Following #417, it is discovered that more methods from `IonReader` could throw an unexpected `AssertionError`. From #417, it is known that `IonReader::stringValue()` which is served by an Amazon implementation of `IonReaderTextSystemX` will throw `AssertionError` when the resolved symbol id is 0 or negative. Although it has been caught by the direct call from the `IonParser::getText()` method, it is found that the call to  `IonReader::next()` from `IonParser.nextToken()` will also invoke `IonReader::stringValue()` in some cases and cause unexpected `AssertionError`.

```java
    @Override
    public JsonToken nextToken() throws IOException
    {
        // special case: if we return field name, we know value type, return it:
        if (_currToken == JsonToken.FIELD_NAME) {
            return (_currToken = _valueToken);
        }
        // also, when starting array/object, need to create new context
        if (_currToken == JsonToken.START_OBJECT) {
            _parsingContext = _parsingContext.createChildObjectContext(-1, -1);
            _reader.stepIn();
        } else if (_currToken == JsonToken.START_ARRAY) {
            _parsingContext = _parsingContext.createChildArrayContext(-1, -1);
            _reader.stepIn();
        }

        // any more tokens in this scope?
        IonType type = null;
        try {
            type = _reader.next();
...
```

The fix is similar to #418, to catch and wrap the `AssertionError`.

We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65273.

PR #433, opened 2023-12-27
Fix for 432: Wrap additional `AssertionError` from `IonReader`

arthurscchan:
This PR fixes https://github.com/FasterXML/jackson-dataformats-binary/issues/432 by wrapping the unexpected AssertionError with the general JsonParseException to avoid throwing AssertionError directly to the users when invalid input is provided.


cowtowncoder:
Thank you @arthurscchan !

PR #435, opened 2023-12-27
Fixed #434: Add null check before accessing integer size

### LINKED ISSUE #434: Unexpected `NullPointerException` thrown from `IonParser::getNumberType()`
In the `IonParser::getNumberType() method`, there is an invocation of the `IonReader.getIntegerSize()` method which could return a `null` value in some cases with invalid data. If the result is null, the code will throw a NullPointerException in the next line when the value is used for the switch condition.

Also, `IonReader.getIntegerSize()` method will throw `NullPointerException` in some cases, thus it is also necessary to wrap around the method invocation to ensure `NullPointerException` is caught.

```java
    public NumberType getNumberType() throws IOException
    {
        IonType type = _reader.getType();
        if (type != null) {
            // Hmmh. Looks like Ion gives little bit looser definition here;
            // harder to pin down exact type. But let's try some checks still.
            switch (type) {
            case DECIMAL:
                //Ion decimals can be arbitrary precision, need to read as big decimal
                return NumberType.BIG_DECIMAL;
            case INT:
                IntegerSize size = _reader.getIntegerSize();
                switch (size) {
...
```
The suggested fix is to add a null checking after the invocation of the `IonReader.getIntegerSize()` method and throw an exception if the return value stored in `size` is indeed null. Also, wrap the `IonReader.getIntegerSize()` method invocation with a try catch block to catch the possible `NullPointerException`.

We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65268 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65274.

arthurscchan:
This PR fixes https://github.com/FasterXML/jackson-dataformats-binary/issues/434 by adding a null check before accessing the result returning from the `IonReader::getIntegerSize()` method and wrap it with a try catch block to handle possible `NullPointerException` thrown from underlying logics.


ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -397,7 +397,15 @@ public NumberType getNumberType() throws IOException
                 //Ion decimals can be arbitrary precision, need to read as big decimal
                 return NumberType.BIG_DECIMAL;
             case INT:
-                IntegerSize size = _reader.getIntegerSize();
+                IntegerSize size = null;
+                try {
+                    size = _reader.getIntegerSize();
+                } catch (NullPointerException e) {
+                    // Possible exception
+                }
```
tgregg, line 405: Note: this try/catch will no longer needed once https://github.com/amazon-ion/ion-java/issues/685 is addressed.

arthurscchan: Thanks for the note. I have added a comments mentioning it is a temporary fix on jackson until amazon-ion/ion-java#685 is addressed and published.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -393,7 +393,17 @@ public NumberType getNumberType() throws IOException
...
                 return NumberType.BIG_DECIMAL;
             case INT:
-                IntegerSize size = _reader.getIntegerSize();
+                IntegerSize size = null;
+                // Temporary measure until this bug fixing is merged and published
+                // https://github.com/amazon-ion/ion-java/issues/685
+                try {
+                    size = _reader.getIntegerSize();
+                } catch (NullPointerException e) {
+                    // Possible exception
```
cowtowncoder, line 402: Hmmh. I assume this is due to invalid/malformer content; and if so, should error out instead of quietly swallowing. Especially since it will fail couple of lines anyway. I can change that.

cowtowncoder (approved):

PR #439, opened 2024-01-02
Add NPE check for dealing with another OSS-Fuzz found edge case

cowtowncoder:


cowtowncoder:
(note: despite branch name, this is for 2.17, not 2.16)

PR #440, opened 2024-01-04
Fix issue reported by OSS-Fuzz (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65513)

cowtowncoder:


arthurscchan:
@cowtowncoder Thanks for picking that up. I just saw this today and I saw that you have fixed that. Besides, there is a new issue coming out yesterday with the same problem (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65628). I assume that will be verified soon with this fix. Thanks again for your help in resolving the OSS-Fuzz reported issue.

PR #441, opened 2024-01-05
Improve error message in AvroAliasTest

apupier:
it gives the reason of schema incompatibility when the schema is expected to be compatible. It can help investigation on #167

difference of output when test is faling (for isntance when upgrading avro dependency to 1.9):
current main branch;
```
[[1;31mERROR[m] [1;31m  AvroAliasTest.testAliasedRecordForwardsCompatible:69 expected:<[]COMPATIBLE> but was:<[IN]COMPATIBLE>
```

with this Pull Request:
```
[[1;31mERROR[m] [1;31m  AvroAliasTest.testAliasedRecordForwardsCompatible:69->checkSchemaIsCompatible:119 Expected schema to be compatible but was not. Reason:
Data encoded using writer schema:
{
  "type" : "record",
  "name" : "Employee",
  "namespace" : "com.fasterxml.jackson.dataformat.avro.AvroTestBase",
  "fields" : [ {
    "name" : "name",
    "type" : "string"
  }, {
    "name" : "age",
    "type" : "int"
  }, {
    "name" : "emails",
    "type" : {
      "type" : "array",
      "items" : "string",
      "java-class" : "[Ljava.lang.String;"
    }
  }, {
    "name" : "boss",
    "type" : "Employee"
  } ]
}
will or may fail to decode using reader schema:
{
  "type" : "record",
  "name" : "NewEmployee",
  "namespace" : "com.fasterxml.jackson.dataformat.avro.interop.annotations.AvroAliasTest",
  "fields" : [ {
    "name" : "name",
    "type" : "string"
  }, {
    "name" : "age",
    "type" : "int"
  }, {
    "name" : "emails",
    "type" : {
      "type" : "array",
      "items" : "string",
      "java-class" : "[Ljava.lang.String;"
    }
  }, {
    "name" : "boss",
    "type" : "NewEmployee"
  } ],
  "aliases" : [ "com.fasterxml.jackson.dataformat.avro.AvroTestBase$.Employee" ]
}
```

### DIFF
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/AvroAliasTest.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/interop/annotations/AvroAliasTest.java
@@ -66,7 +66,7 @@ public class AvroAliasTest extends InteropTestBase {
         SchemaCompatibility.SchemaPairCompatibility compatibility =
             SchemaCompatibility.checkReaderWriterCompatibility(newSchema, oldSchema);
         //
-        assertThat(compatibility.getType()).isEqualTo(SchemaCompatibility.SchemaCompatibilityType.COMPATIBLE);
+        checkSchemaIsCompatible(compatibility);
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


cowtowncoder (approved):

cowtowncoder:
Makes sense, thanks!

PR #442, opened 2024-01-05
Use valid schema related to null default based on specification in

apupier:
RecordEvolutionTest

during investigation upgrading Avro to 1.9+, the test was failing. the code has enforced the rule from the specification that `when a default value is specified for a record field whose type is a union, the type of the default value must match the first element of the union. Thus, for unions containing "null", the "null" is usually listed first, since the default value of such unions is typically null.`

it was part of specification in 1.8 already
https://avro.apache.org/docs/1.8.0/spec.html#Unions , just not enforced in the codebase.

relates to #167


cowtowncoder (approved):

cowtowncoder:
Ah! Thank you, yes this is interesting.

Issue #449, opened 2024-01-08
`IndexOutOfBoundsException` in `JacksonAvroParserImpl` for invalid input


arthurscchan:
In the `JacksonAvroParserImpl::_finishShortText(int)` method, there is a missing bound check-in handling the value reading from the byte array `inputBuf` and could cause unexpected `IndexOutOfBoundsException` if the provided input is not correctly ended.

In the first line of the `do..while` loop, the current index pointed by the `inPtr` variable is retrieved, processed and stored. Then the `inPtr` value is increased by one. From the study of the code, the value of `inPtr` must be within the range of the byte array `inputBuf`, but since it has increased by one in this line, the value of `inPtr` may be larger or equals to the length of `inputBuf` on some invalid input. This causes the subsequent `inputBuf` value access with the out-of-bound `inPtr` to throw an unexpected `IndexOutOfBoundsException`.

```java
...
        final int[] codes = sUtf8UnitLengths;
        do {
            i = inputBuf[inPtr++] & 0xFF;
            switch (codes[i]) {
            case 0:
                break;
            case 1:
                i = ((i & 0x1F) << 6) | (inputBuf[inPtr++] & 0x3F);
                break;
            case 2:
                i = ((i & 0x0F) << 12)
                   | ((inputBuf[inPtr++] & 0x3F) << 6)
                   | (inputBuf[inPtr++] & 0x3F);
                break;
            case 3:
                i = ((i & 0x07) << 18)
                 | ((inputBuf[inPtr++] & 0x3F) << 12)
                 | ((inputBuf[inPtr++] & 0x3F) << 6)
                 | (inputBuf[inPtr++] & 0x3F);
                // note: this is the codepoint value; need to split, too
                i -= 0x10000;
                outBuf[outPtr++] = (char) (0xD800 | (i >> 10));
                i = 0xDC00 | (i & 0x3FF);
                break;
            default: // invalid
                _reportError("Invalid byte "+Integer.toHexString(i)+" in Unicode text block");
            }
            outBuf[outPtr++] = (char) i;
        } while (inPtr < end);
        return _textBuffer.setCurrentAndReturn(outPtr);
...
```
Similar problem happened in `JacksonAvroParserImpl::_finishLongText(int)` method.

```java
    private final void _finishLongText(int len) throws IOException
...
            case 3: // 4-byte UTF
                c = _decodeUTF8_4(c);
                // Let's add first part right away:
[...]

PR #450, opened 2024-01-08
Fixes for 449: Unexpected IOOBE handling

arthurscchan:
This PR provides a possible fix for #449 by adding a bound check to avoid unexpected IOOBE from avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/JacksonAvroParserImpl.java.


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/JacksonAvroParserImpl.java
```diff
@@ -628,6 +628,9 @@ private final String _finishShortText(int len) throws IOException
         final int[] codes = sUtf8UnitLengths;
         do {
             i = inputBuf[inPtr++] & 0xFF;
+            if (inPtr >= inputBuf.length) {
```
cowtowncoder, line 631: Not quite correct: need to compare to `end`, not physical buffer end (two may or may not be same).

But I can work with this, thanks!

cowtowncoder (approved):

cowtowncoder:
Thank you @arthurscchan !

Issue #451, opened 2024-01-08
`IndexOutOfBoundsException` in `CBORParser` for invalid input


arthurscchan:
The `CBORParser::nextToken()` method relies on the integer index `_inputPtr ` to read the next character from the provided input byte array. In some cases, if the provided input byte array is malformed and contains negative bytes, that negative could be used as the new value for the `_inputPtr`. If the negative `_inputPtr` is used as an index for later access to the byte array, an unexpected `IndexOutOfBoundsException` is thrown because a negative index is used.

```java
    @Override
    public JsonToken nextToken() throws IOException
    {
...
        if (_inputPtr >= _inputEnd) {
            if (!loadMore()) {
                return _eofAsNextToken();
            }
        }
        int ch = _inputBuffer[_inputPtr++] & 0xFF;
...
```
The suggested fix is to add a negative checking before the use of `_inputPtr`. It is shown that there is already a check in the method to ensure `_inputPtr` is not larger than or equal to the `_inputEnd`, but there is no check to confirm that `_inputPtr` is not negative. The suggested fix is to add a negative check to ensure the retrieved `_inputPtr` is not negative before use.

We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65617.

PR #452, opened 2024-01-08
Fix for 451: Add negative index checking for token retrieval

arthurscchan:
This PR provides a suggested fix for #451 by adding a negative index checking in `CBORParser` to ensure that the index `_inputPtr` is not negative in additional to the original check to ensure it is not larger than or equals to the length of the input.


cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -813,6 +813,9 @@ public JsonToken nextToken() throws IOException
                 return _eofAsNextToken();
             }
         }
+        if (_inputPtr < 0) {
```
cowtowncoder, line 816: I don't think this is a very useful check: we must fix incorrect input pointer update where it occurs -- it is possible that negative offset was used to move back within buffer, but without becoming negative.

cowtowncoder:
Thank you again, @arthurscchan ! Modified fix a bit, should resolve the issue.

PR #453, opened 2024-01-09
Additional fix for issue 449

arthurscchan:
This PR suggests a fix for #449 in additional to the fix in #450.


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/deser/JacksonAvroParserImpl.java
```diff
@@ -700,6 +700,12 @@ private final void _finishLongText(int len) throws IOException
                 break;
             case 3: // 4-byte UTF
                 c = _decodeUTF8_4(c);
+                // if outPtr >= outBuf.length before adding
+                // the input is malformed.
```
cowtowncoder, line 704: Actually, not necessarily so. But we do need to check there's room. I'll tweak this PR a bit.

cowtowncoder:
Thank you again @arthurscchan ! Very quick resolution again :)

PR #454, opened 2024-01-11
Fix another OSS-Fuzz identified AIOOB exception wrt timestamps

cowtowncoder:


cowtowncoder:
New unit test failure not due to this PR but probably recent fix to Enum field renaming by jackson-databind; merging.

Issue #458, opened 2024-01-16
Unexpected NullPointerException in `CBORParser`


arthurscchan:
In the `CBORParser.convertNumberToBigDecimal()` method, there is an invocation of the `CBORParser.getText()` method which could return a `null` value when there is no more text left in the input. If the result is null, the code will throw a NullPointerException in the next line when the `String::length()` method is called. The `CBORParser.convertNumberToBigDecimal()` method is called by the public API `CBORParser::nextDecimalValue()`. 

```java
    @Override
    public BigDecimal getDecimalValue() throws IOException
    {
        if ((_numTypesValid & NR_BIGDECIMAL) == 0) {
            if (_numTypesValid == NR_UNKNOWN) {
                _checkNumericValue(NR_BIGDECIMAL);
            }
            if ((_numTypesValid & NR_BIGDECIMAL) == 0) {
                convertNumberToBigDecimal();
            }
        }
        return _numberBigDecimal;
    }
```
```java
    protected void convertNumberToBigDecimal() throws IOException
    {
        // Note: this MUST start with more accurate representations, since we don't know which
        //  value is the original one (others get generated when requested)
        if ((_numTypesValid & (NR_DOUBLE | NR_FLOAT)) != 0) {
            // Let's parse from String representation, to avoid rounding errors that
            //non-decimal floating operations would incur
            final String text = getText();
            streamReadConstraints().validateFPLength(text.length());
...
```

The suggested fix is to add a null checking after the invocation of the `ICBORParser.getText()` method and throw an exception if the return value stored in `size` is indeed null. 

We found this issue by OSS-Fuzz and it is reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65768.

PR #459, opened 2024-01-16
Fix issue 458: Add null checking

arthurscchan:
This PR fixes issue #458 by adding a null check before accessing the length of the text string. This could avoid errors when getDecimalValue is invoked when there is no more data left in the input.


cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -2224,6 +2224,9 @@ protected void convertNumberToBigDecimal() throws IOException
             // Let's parse from String representation, to avoid rounding errors that
             //non-decimal floating operations would incur
             final String text = getText();
+            if (text == null) {
+                throw _constructReadException("No more values");
```
cowtowncoder, line 2228: Ok; I'll see what exactly is going on here.

cowtowncoder (approved):
With minor change to fix, LGTM, approved, will merge.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -1112,6 +1112,9 @@ protected JsonToken _handleTaggedBinary(TagList tags) throws IOException
         } else {
             // 12-May-2016, tatu: Since that's all we know, let's otherwise
             //   just return default Binary data marker
+            // 16-Jan-2024, tatu: Esoteric edge case where we have marked
+            //   `int` as being tokenized
+            _numTypesValid = NR_UNKNOWN;
```
cowtowncoder, line 1117: This is the actual fix: for some reason `int` tag indicated number token. May need to revisit logic at some point but this should work for now.

arthurscchan:
Thanks for showing me the details of the problematic part.

PR #468, opened 2024-01-24
Upgrade ion-java to 1.11.1 and remove catch clauses for exceptions that are no longer leaked.

tgregg:
ion-java 1.11.1 fixed several cases where unchecked exceptions were leaking and had to be handled with IonParser. This pull request upgrades the ion-java dependency to 1.11.1 and cleans up the catch clauses in IonParser.

I've added two comments in the code to point to https://github.com/amazon-ion/ion-java/issues/702 , which will address a case where AssertionError is still being leaked. I will submit another pull request here once that is fixed.

I tried, unsuccessfully, to reproduce a NegativeArraySizeException in `IonReader.newBytes()`, as was indicated by OSS-Fuzz#66077, which I'm not sure if I'm able to access. If you still have the reproduction for this, let me know. Otherwise I propose we remove the catch and see if OSS-Fuzz finds this issue again. It's possible that other fixes in 1.11.1 fixed this as well, but I'm not sure. I'm more confident about the others, though there's still a chance that something that isn't covered in tests will pop up again on the report. I'll be ready to respond if that happens.

### DIFF
--- a/ion/pom.xml
+++ b/ion/pom.xml
@@ -39,7 +39,7 @@ tree model)
     <dependency>
       <groupId>com.amazon.ion</groupId>
       <artifactId>ion-java</artifactId>
-      <version>1.11.0</version>
+      <version>1.11.1</version>
     </dependency>
 
     <!-- Extends Jackson core, databind -->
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):
LGTM!

cowtowncoder:
@tgregg That makes sense: I am ok in removing checks, seeing what can be found.

So, will go ahead and merge changes as suggested; `AssertionError`s can be tackled in follow-up pr.

Issue #469, opened 2024-01-25
IonReader.newBytes() throwing `NegativeArraySizeException`


cowtowncoder:
OSS-Fuzz reported a problem with stack trace of:

```
== Java Exception: java.lang.NegativeArraySizeException: -81
	 at com.amazon.ion.impl.IonReaderContinuableCoreBinary.newBytes(IonReaderContinuableCoreBinary.java:565)
	 at com.fasterxml.jackson.dataformat.ion.IonParser.getEmbeddedObject(IonParser.java:592)
	 at com.fasterxml.jackson.databind.deser.std.BaseNodeDeserializer._fromEmbedded(JsonNodeDeserializer.java:836)
	 at com.fasterxml.jackson.databind.deser.std.BaseNodeDeserializer._deserializeRareScalar(JsonNodeDeserializer.java:697)
	 at com.fasterxml.jackson.databind.deser.std.BaseNodeDeserializer._deserializeContainerNoRecursion(JsonNodeDeserializer.java:645)
	 at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:104)
	 at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:25)
	 at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:342)
	 at com.fasterxml.jackson.databind.ObjectMapper._readValue(ObjectMapper.java:4876)
	 at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:3145)
```

with specifically generated corrupt Ion document.

cowtowncoder:
cc @tgregg -- this is first of 3 new/re-created OSS-Fuzz issues: will add PR for reproduction.

PR #470, opened 2024-01-25
Add test for #469 under `failing` (to avoid breaking CI)

cowtowncoder:


cowtowncoder:
@tgregg Here's failing test for #469

Issue #473, opened 2024-01-25
`IonReader.next()` throws `ArrayIndexOutOfBoundsException` for some corrupt content


cowtowncoder:
(note: OSS-Fuzz sourced from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66131)

Looks like there's a code path that produces AIOOBE for corrupt content, with stack trace of:

```
 	== Java Exception: java.lang.ArrayIndexOutOfBoundsException: Index 16 out of bounds for length 16
	 at com.amazon.ion.impl.IonCursorBinary.uncheckedNextToken(IonCursorBinary.java:1571)
	 at com.amazon.ion.impl.IonCursorBinary.nextValue(IonCursorBinary.java:1714)
	 at com.amazon.ion.impl.IonReaderContinuableCoreBinary.nextValue(IonReaderContinuableCoreBinary.java:460)
	 at com.amazon.ion.impl.IonReaderContinuableApplicationBinary.nextValue(IonReaderContinuableApplicationBinary.java:942)
	 at com.amazon.ion.impl.IonReaderContinuableTopLevelBinary.handleIncompleteValue(IonReaderContinuableTopLevelBinary.java:150)
	 at com.amazon.ion.impl.IonReaderContinuableTopLevelBinary.next(IonReaderContinuableTopLevelBinary.java:166)
	 at com.fasterxml.jackson.dataformat.ion.IonParser.nextToken(IonParser.java:657)
	 at com.fasterxml.jackson.core.JsonParser.nextTextValue(JsonParser.java:1161)

```

cowtowncoder:
Unable to reproduce at this point; most code paths / call sequences produce proper `IonException`.
Will try to tweak test case bit more to reproduce.

cowtowncoder:
@tgregg This is for the 3rd OSS-Fuzz reported new-or-recycled issue, I was unable to reproduce.

@arthurscchan Maybe you can reproduce this issue? Idea is to reproduce but not add defensive programming yet, to let `ion-java` developers to handle at appropriate level

tgregg:
#477 tweaks the test case to reproduce the problematic stack trace.

cowtowncoder:
Test now fails properly. Thank you @tgregg !

PR #474, opened 2024-01-25
Add not-failing test that tries to reproduce #473: has the data but call sequence not same as OSS-Fuzzer

cowtowncoder:
So, Work-in-Progress test that should reproduce #473 but does not yet.


cowtowncoder:
@arthurscchan @tgregg Here's the not-yet-failing test: Ion doc from OSS-Fuzz, just need right call sequence to reproduce (failure is for `IonParser.nextTextValue()`)

PR #477, opened 2024-01-27
Update IonFuzz_473_66131_AIOOBE_Test to provoke an ArrayIndexOutOfBoundsException

tgregg:
This PR proposes to update draft PR #474 so that it produces the same exception trace as the original OSS-Fuzz report. This test is still failing for now (with `ArrayIndexOutOfBoundsException`), but I confirmed it passes with the fix from https://github.com/amazon-ion/ion-java/pull/710. Once that lands in the next version of ion-java I will open a PR to move this test out of `failing`.

Note: this can also be reproduced minimally with the data following data: `e0 01 00 ea dc 9a`.


cowtowncoder (approved):

cowtowncoder:
Ok, I think I'll try the minimal reproduction then. Thanks!

cowtowncoder:
Thank you @tgregg !

Issue #428, opened 2023-12-20
`IonParser.getIntValue()` fails or does not handle value overflow checks


cowtowncoder:
(note: found via https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65180)

Looks like overflow checks not being by `IonParser` can result in one of outcomes:

* an Ion `RuntimeException` being thrown, when method `getIntValue()` is called with value beyond 32-bit `int` range
* quiet truncation of value without failure

Instead, this should be surfaced same as equivalent `JsonParseException` JSON-backed `JsonParser` throws in similar situation -- see `ParserBase.convertNumberToInt()` (and `ParserBase.reportOverflowInt()`) for examples of handling.

Same probably affects `long` overflow as well (wrt `BigInteger` range value)

cowtowncoder:
Since I do not have time to work on adding coercion checks quite yet, will just add failing tests for future work.
Should be relatively straight-forward for someone to pick up.

PR #481, opened 2024-02-09
Handle value overflow checks for IonParser.getIntValue() - #428

thomasdelange5:
Add number size checks for IonParser getIntValue and getLongValue. Fix failing tests for #428


ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -390,8 +390,25 @@ public int getIntValue() throws IOException {
...
+                BigInteger bigInteger = _reader.bigIntegerValue();
+                if (BI_MIN_INT.compareTo(bigInteger) > 0 || BI_MAX_INT.compareTo(bigInteger) < 0) {
+                    this.reportOverflowInt();
+                }
+                return bigInteger.intValue();
+            } else {
+                return _reader.intValue();
+            }
+        } catch (IonException
+                // 15-Jan-2024, tatu: other OSS-Fuzz tests suggest we need this:
```
cowtowncoder, line 410: Let's not add this back (I think @tgregg is working on handling these edge cases)

cowtowncoder:
Looks good @thomasdelange5 -- thank you for contributing this!

Only one other thing before I can merge this: if we haven't gotten CLA for you yet (it only needs to be done once per contributor):

(or possibly alternate Corporate CLA addendum if this is under some company's; we have a few Amazon contributors for Ion).
It is usually easiest to fill & sign, print, scan/photo, email to `cla` at fasterxml dot com.
Once we have it (and if you have sent previously, let me know please), I can proceed with merging this fix!

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
tgregg: Correct. If fuzzing identifies any new leaked exceptions I'd like to fix those in ion-java; in the meantime we can cover them in `failing` tests.

thomasdelange5:
Thank you for getting back so quickly @cowtowncoder! I will remove that catch clause and submit the CLA now.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -390,7 +390,22 @@ public int getIntValue() throws IOException {
     // @since 2.17
     private int _getIntValue() throws IOException {
         try {
-            return _reader.intValue();
+            if (getNumberType() == NumberType.LONG) {
```
tgregg, line 393: Minor: for performance reasons, I recommend storing this value and using it in both branches. ion-java has to do some bounds checking on the values, and we should avoid doing that twice when possible.

thomasdelange5: Ok will do. Thanks @tgregg

cowtowncoder: was about to suggest same, +1

tgregg (approved):

cowtowncoder (approved):

cowtowncoder:
CLA received, will merge for 2.17(.0).

Thank you, @thomasdelange5 !

thomasdelange5:
Awesome, thank you very much!

PR #484, opened 2024-03-27
Rework synchronization in `ProtobufMapper`

pjfanning:
* also marked the variable as volatile as this is something that is often done when you have concurrent access
* the new code tries a little harder not to create multiple DescriptorLoaders. The old code could have queued up multiple threads where l was null and then each got the synchronization lock in turn and then created a fresh DescriptorLoader.


pjfanning:
@cowtowncoder would you have time to review this?

cowtowncoder (approved):
LGTM: will rename lock member, add release notes.

cowtowncoder:
> @cowtowncoder would you have time to review this?

Will do, thank you for reminder.

PR #496, opened 2024-05-19
[Avro] write enums as string

### LINKED ISSUE #494: Avro Schema generation: allow mapping Java Enum properties to Avro String values
With json, I could always extend an enum (on both sender and receiver ends), consider it a backwards compatible change,  and the receiver could parse both old and new documents equally well.

With avro, altering an enum makes the entire schema incompatible, and an updated schema can not be used to read old docs.
For all my use cases, serialising all enums as string would be a completely acceptable strategy for avoiding this problem, and much preferable to versioned formats,
but jacksons avro support does not currently seem to offer this.  

So I'm hoping for a new feature switch for avro: WRITE_ENUMS_AS_STRINGS or similar.
 
In my case I only really care about serialising, since the receiver is on a completely different platform, but I guess parity on the deserialising side would be natural to include.

MichalFoksa:
New switch added `enableWriteEnumAsString()` to enable Java enum to Avro string mapping at schema generation.

Fixes #494

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaGenerator.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaGenerator.java
@@ -31,4 +31,17 @@ public class AvroSchemaGenerator extends VisitorFormatWrapperImpl
         super.disableLogicalTypes();
         return this;
     }
+
+    @Override
+    public AvroSchemaGenerator enableWriteEnumAsString() {
+        super.enableWriteEnumAsString();
+        return this;
+    }
+
+    @Override
+    public AvroSchemaGenerator disableWriteEnumAsString() {
+        super.disableWriteEnumAsString();
+        return this;
+    }
+
 }
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/EnumVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/EnumVisitor.java
@@ -1,0 +1,45 @@
+package com.fasterxml.jackson.dataformat.avro.schema;
+
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonStringFormatVisitor;
+
+import org.apache.avro.Schema;
+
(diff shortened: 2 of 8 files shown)


cowtowncoder:
Thank you @MichalFoksa ! PR very welcome.

I hope to check this out soon -- possibly not today but in next couple of days.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/AvroSchemaGenerator.java
```diff
@@ -31,4 +31,17 @@ public AvroSchemaGenerator disableLogicalTypes() {
         super.disableLogicalTypes();
         return this;
     }
+
+    @Override
+    public AvroSchemaGenerator enableWriteEnumAsString() {
```
cowtowncoder, line 36: Quick note: it'd be better to add a new `AvroGenerator.Feature` entry instead of new method(s), if possible.
Would that work?

MichalFoksa: > Yes, AvroGenerator.Feature would be better but I'd need to think of plumbing.
>
> I think AvroGenerator might be available from SerializerProvider but not sure if that is properly initialized.

ATM I do not know how to access `AvroGenerator.Feature` from `VisitorFormatWrapperImpl.expectStringFormat(JavaType)`.

MichalFoksa: I did not find a way to read `AvroFactory`, or how to elegantly pass it into `VisitorFormatWrapperImpl`. :(

cowtowncoder: Ok I hope to have today or tomorrow to see if I can figure out something...

cowtowncoder: @MichalFoksa Answer would `SerializerProvider` which is assigned to `VisitorFormatWrapperImpl` -- it has `getGenerator()` method. Alas, I am not convinced provider gets assigned said generator for this case.

So I'll probably accept this as-is -- let me do final code review and proceed.

cowtowncoder (approved):

cowtowncoder:
@MichalFoksa I think this is ready so will merge; made some minor cosmetic changes but nothing major.

Thank you once again for contributing this!

PR #512, opened 2024-08-21
Update avro to 1.11.3 (backport to 2.18 branch)

rafalh:
Update Apache avro library to 1.11.3.
To make it compatible namespace generation for nested classes had to be changed.
Now dots are used instead of dollar signs as indicator of nesting.
AFAIK dollar sign is not an allowed character in Avro namespace according to the specification and this is why this change was implemented in avro library.
Note that avro library generates namespaces not ending with dollar character since version 1.9, but it actually removes all dollar signs in classes with multiple levels of nesting since 1.11.

See: AVRO-2143 and AVRO-2757
Fixes https://github.com/FasterXML/jackson-dataformats-binary/issues/167

Compatibility:
* over-the-wire protocol should be compatible
* reading schemas from older versions should work
* schemas generated after this change for classes with no nesting are unchanged
* schemas generated after this change for classes with one nesting level are readable by Jackson 2.11+ (commit c5705492)
* schemas generated after this change for classes with multiple nesting levels will not be readable by older Jackson versions

### DIFF
--- a/avro/pom.xml
+++ b/avro/pom.xml
@@ -47,7 +47,7 @@ abstractions.
     <dependency>
       <groupId>org.apache.avro</groupId>
       <artifactId>avro</artifactId>
-      <version>1.8.2</version>
+      <version>1.11.3</version>
     </dependency>
 
     <!-- and for testing we need logback -->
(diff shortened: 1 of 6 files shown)


cowtowncoder:
Quick note: as per my comments, happy to review, help get merged.
But one thing we'll need before merging, CLA, from:

It is only needed once before the first contribution, and the usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once I get that I can proceed with final review and merging.
Thank you in advance, looking forward to FINALLY resolving the upgrade problem!

rafalh:
I've sent CLA photo. I just noticed that I used `info` e-mail instead of `cla`, because that's what was used in [CONTRIBUTING.md](https://github.com/FasterXML/jackson/blob/master/CONTRIBUTING.md). Thanks for letting me know.

cowtowncoder:
@rafalh `info` is fine too; will need to change `CONTRIBUTING.md`

cowtowncoder:
CLA received, will do one more review.

cowtowncoder:
Thank you again, @rafalh -- this is awesome! Hoping to wrap up 2.18.0-rc1 planning Very Soon Now, to get the release candidate out.

PR #539, opened 2024-12-27
Update `avro` dep from 1.11.3 to 1.11.4

cowtowncoder:


cowtowncoder:
Fails with 

```
[ERROR]   StringableTest.testCustomStringableArray:193->InteropTestBase.roundTrip:151 » Security Forbidden class com.fasterxml.jackson.dataformat.avro.interop.annotations.StringableTest$CustomStringableKey! This class is not trusted to be included in Avro schema using java-class. Please set org.apache.avro.SERIALIZABLE_PACKAGES system property with the packages you trust.
```

@MichalFoksa I hope there is some other way to do this than Global system properties...

cowtowncoder:
Avro lib's compatibility is once again anything but SemVer -- failing existing use cases with PATCH change is not... good.

Be that as it may, changing target branch to 2.19 since I don't think patch should have such incompatible change.

MichalFoksa:
Hmm it is cause of [AVRO-3985 - Restrict trusted packages in ReflectData and SpecificData](https://issues.apache.org/jira/browse/AVRO-3985) and [CVE-2024-47561](https://avd.aquasec.com/nvd/2024/cve-2024-47561/).

MichalFoksa:
To fix test add `com.fasterxml.jackson.dataformat.avro.interop.annotations` in to "serializable packages" property.

```
    @Before
    public void init() {
        System.setProperty("org.apache.avro.SERIALIZABLE_PACKAGES",
                "java.lang,java.math,java.io,java.net,org.apache.avro.reflect," + // These are default trusted packages by Avro 1.11.4
                        "com.fasterxml.jackson.dataformat.avro.interop.annotations");
    }

PR #540, opened 2024-12-30
SmileParser getValueAsString() issue with JsonToken.FIELD_NAME

johhud1:
Test updates to reproduce the issue and naive fix attempt for #541

### DIFF
--- a/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/POJOSimpleReadTest.java
+++ b/avro/src/test/java/com/fasterxml/jackson/dataformat/avro/POJOSimpleReadTest.java
@@ -60,6 +60,8 @@ public class POJOSimpleReadTest extends AvroTestBase
         assertFalse(p.hasTextCharacters());
         assertEquals("name", p.currentName());
         assertEquals("name", p.getText());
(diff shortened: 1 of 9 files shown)


cowtowncoder:
Ok I'll go over the PR; I think it makes sense overall.

But one process thing we need before I can merge the fix (assuming review goes fine); unless already done, we'd need a CLA. It's here:

and only needs to be sent one before the first contribution is merged (good for all future PRs).
The usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once I get it can proceed with merging (and probably in this case also backporting in 2.18 at least, maybe 2.17)

johhud1:
Just emailed my signed CLA

cowtowncoder:
CLA received.

cowtowncoder (approved):
Looks good; will extend a bit to handle other formats too.

cowtowncoder:
Ok: added tests to ensure all 5 format backends work (as well as test for `JsonParser.getValueAsString(String defaultValue)`) -- and matching fixes (some worked, some didn't).

Thank you for reporting this @johhud1 and submitting a good fix for Smile case.

Issue #547, opened 2025-01-11
JSTEP-10: Unify testing structure/tools


JooHyukKim:
This issue serves as an umbrella issue for resolving [[JSTEP-10: Unifying testing structure/tools]](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-10)

### TODO

- Add JUnit 5 dep + `failing` test utils
    - #548 
- Migrate modules to JUnit 5
    - #549 
    - #550 
    - #551 
    - #552 
    - #553 
- Migrate failing tests to `/tofix` and remove existing `/failing` funciontality
    - #554 
- Remove JUnit 4 dependecy (isolate to `avro` module's `/interop` tests at least) 
    - #555
- Migrate `/interop` tests from `avro` module to JUnit 5

### How to do things

Change Code

- run IntelliJ's "migrate tests 4 -> 5"
- Check for Junit 3 TestCase and remove the extends part
- Annotate `@Test` by : Replace (by text) all `public void test` to `@Test \n public void test`
- Add imports by : Replace `public class` with `import org.jun... Test \n import static Asssertions.*; \n public class`

Validate
- search all `public void test...` and check Dups or missing(indents)
- search all `public class...` and check JavaDoc in between imports
- search and remove all `org.junit.Assert`
- search and remove `Assert.*`
- search all `assertNotNull` error message String usage

JooHyukKim:
@cowtowncoder Could you open this one up again? I made a mistake above PRs resolve this.

Didn't know "partly resolves #xxx" would also do this 😅

cowtowncoder:
Yeah, github is being too "smart". :-(

cowtowncoder:
Completed, I think!

PR #553, opened 2025-01-12
[ JSTEP-10 ] Migrate `smile` module tests to JUnit 5 

JooHyukKim:
related #547

Blocked by #548


JooHyukKim (commented):
Looks ok

smile/src/test/java/com/fasterxml/jackson/dataformat/smile/VersionsTest.java
```diff
@@ -1,14 +1,17 @@
 package com.fasterxml.jackson.dataformat.smile;
 
-import java.io.*;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
```
JooHyukKim, line 4: @cowtowncoder One thing tho, IntelliJ doesn't provide per-package wildcard handling 🤔
But with that trade off, we get to order imports Jackson way

image

JooHyukKim: image

cowtowncoder: I think that ideally we'd usually left imports as-is and avoid mechanical changes.

But FWTW, the usual cut-off is 5-or-more, I think. Then again for some packages (`java.io`, `java.util`) could use star-include for lower numbers.

I think that order of imports is more important than exact threshold for start includes, if that helps? (i.e. I'd rather have strict(er) ordering than worry about explicit vs star inclusion, if only one can be restricted).

cowtowncoder (approved):

PR #573, opened 2025-04-09
Fix #571: Unable to deserialize a pojo with IonStruct

### LINKED ISSUE #571: Unable to deserialize a pojo with IonStruct
I am not seeing a good option to serialize/deserialize a simple Pojo with a `null` IonStruct. 

The issue is that by default a null IonStruct gets serialized to `null` but due to this change https://github.com/FasterXML/jackson-dataformats-binary/pull/296 that gets converted to a an IonNull, and fails with argument type mismatch.

Failing testing:

```
  public static class Pojo {    
        @JsonCreator
        public Pojo(@JsonProperty("name") String name, @JsonProperty("data") IonStruct data) {   
            this.name = name;
            this.data = data;
        }
        public String name;
        public IonStruct data;
        
        
        @Override
        public int hashCode() {
            return Objects.hash(data, name);
        }
        @Override
        public boolean equals(Object obj) {
            if (this == obj)
                return true;
            if (obj == null)
                return false;
            if (getClass() != obj.getClass())
                return false;
            Pojo other = (Pojo) obj;
            return Objects.equals(data, other.data) && Objects.equals(name, other.name);
        }
    }

    private final IonSystem ionSystem = IonSystemBuilder.standard().build();
    private final IonValueMapper mapper = new IonValueMapper(ionSystem);

    @Test
    public void testRoundTrip() throws JsonProcessingException {
        Pojo p = new Pojo("test", null);
        String value =  mapper.writeValueAsString(p);
        Assert.assertEquals("{name:\"test\",data:null}", value);
        Pojo p2 = mapper.readValue(value, Pojo.class);
        Assert.assertEquals(p, p2);
    }
```

seadbrane:
(fixes #571)

Make IonValueDeserializer a ContextualDeserializer in order to create the correct Ion null type.   

Also, add new IonParser feature READ_NULL_AS_IONVALUE in order to be able to serialize `null` as either an IonNull or java `null`.

### DIFF
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -53,6 +53,18 @@ public class IonParser
          * @since 2.12.3
          */
         USE_NATIVE_TYPE_ID(true),
+        /**
+         * Whether to convert "null" to an IonValueNull (true);
+         * or leave as a java null (false) when deserializing.
+         *<p>
+         * Enabled by default for backwards compatibility as that has been the behavior
+         * of `jackson-dataformat-ion` since 2.13.
+         *
+         * @see <a href="https://amzn.github.io/ion-docs/docs/spec.html#annot">The Ion Specification</a>
+         *
+         * @since 2.19.0
+         */
+        READ_NULL_AS_IONVALUE(true),
         ;
 
         final boolean _defaultState;
@@ -563,6 +575,12 @@ public class IonParser
         writer.writeValue(_reader);
         IonValue v = l.get(0);
         v.removeFromContainer();
+
+        if (v.isNullValue() && !Feature.READ_NULL_AS_IONVALUE.enabledIn(_formatFeatures)) {
+            if (_valueToken == JsonToken.VALUE_NULL && !IonType.isContainer(v.getType())) {
+                return null;
+            }
+        }
         return v;
     }
 
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
@@ -576,8 +576,13 @@ public class IonParser
         IonValue v = l.get(0);
         v.removeFromContainer();
 
-        if (v.isNullValue() && !Feature.READ_NULL_AS_IONVALUE.enabledIn(_formatFeatures)) {
-            if (_valueToken == JsonToken.VALUE_NULL && !IonType.isContainer(v.getType())) {
+        if (!Feature.READ_NULL_AS_IONVALUE.enabledIn(_formatFeatures)) {
+            // 2025-04-11, seadbrane: The default is to read 'null' as an Ion Null object.
+            // However, there is no way to determine from the serialized ion data if a 'null'
+            // was an IonNullValue or a 'null' container type such as IonNullStruct or IonNullList.
+            // So if READ_NULL_AS_IONVALUE is disabled, then return 'null' if the _valueToken
+            // is 'null' and the Ion value read is not container type already.
+            if (v.isNullValue() && _valueToken == JsonToken.VALUE_NULL && !IonType.isContainer(v.getType())) {
                 return null;
             }
         }
--- a/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
+++ b/ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
@@ -16,21 +16,51 @@ package com.fasterxml.jackson.dataformat.ion.ionvalue;
 
 import java.io.IOException;
 
-import com.amazon.ion.*;
-
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
-import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.BeanProperty;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.deser.ContextualDeserializer;
+import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.util.AccessPattern;
 import com.fasterxml.jackson.dataformat.ion.IonParser;
+import com.amazon.ion.IonContainer;
+import com.amazon.ion.IonList;
+import com.amazon.ion.IonSexp;
+import com.amazon.ion.IonStruct;
+import com.amazon.ion.IonSystem;
+import com.amazon.ion.IonType;
+import com.amazon.ion.IonValue;
+import com.amazon.ion.Timestamp;
 
 /**
  * Deserializer that knows how to deserialize an IonValue.
  */
-class IonValueDeserializer extends JsonDeserializer<IonValue> {
+class IonValueDeserializer extends JsonDeserializer<IonValue> implements ContextualDeserializer {
+
+    private final JavaType targetType;
+
+    public IonValueDeserializer() {
+        this.targetType = null;
+    }
+
+    public IonValueDeserializer(JavaType targetType) {
+        this.targetType = targetType;
+    }
+
+    @Override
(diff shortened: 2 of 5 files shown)


ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
```diff
@@ -563,6 +575,12 @@ private IonValue getIonValue() throws IOException {
         writer.writeValue(_reader);
         IonValue v = l.get(0);
         v.removeFromContainer();
+
+        if (v.isNullValue() && !Feature.READ_NULL_AS_IONVALUE.enabledIn(_formatFeatures)) {
```
cowtowncoder: I know existing is the way it is, but call above to:

    _updateToken(JsonToken.VALUE_EMBEDDED_OBJECT);

looks wrong, given this is called (only) from `getEmbeddedObject()`... which should only be called when we do have `JsonToken.VALUE_EMBEDDED_OBJECT` (and if not, fail).

Would probably make sense to add a comment here to explain logic wrt `!IonType.isContainer(v.getType())` -- I am not sure I understand the logic myself.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
```diff
@@ -16,21 +16,51 @@
...
+import com.amazon.ion.IonValue;
+import com.amazon.ion.Timestamp;
 
 /**
  * Deserializer that knows how to deserialize an IonValue.
  */
-class IonValueDeserializer extends JsonDeserializer<IonValue> {
+class IonValueDeserializer extends JsonDeserializer<IonValue> implements ContextualDeserializer {
+
+    private final JavaType targetType;
```
cowtowncoder: Member variables with leading underscore -> `_targetType`

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
```diff
@@ -16,21 +16,51 @@
 
 import java.io.IOException;
 
-import com.amazon.ion.*;
-
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
-import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.BeanProperty;
```
cowtowncoder: Usually fold imports of 5+ from same package.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
```diff
@@ -62,17 +92,34 @@ public IonValue getNullValue(DeserializationContext ctxt) throws JsonMappingExce
...
+                        IonType containerType = getIonContainerType();
+                        if (containerType != null) {
+                            IonSystem ionSystem = ((IonParser) parser).getIonSystem();
+                            return ionSystem.newNull(containerType);
+                        }
                         return iv;
                     }
                 }
             }
-
```
cowtowncoder, line 69: What we also need, to make sure `getNullValue()` call is made for every `null` is override of

```
    @Override
    public AccessPattern getNullAccessPattern() {
        // Default implementation assumes that the null value does not vary, which
        // is usually the case for most implementations. But it is not necessarily
        // `null`; so sub-classes may want to refine further.
        return AccessPattern.CONSTANT;
    }
```

to instead return `AccessPattern.DYNAMIC`: otherwise value returned on first call is cached by parent deserializer.

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
```diff
@@ -65,7 +73,7 @@ static class IonValueData extends Data<IonValue> {
     }
 
     private static final IonSystem SYSTEM = IonSystemBuilder.standard().build();
-    private static final IonValueMapper ION_VALUE_MAPPER = new IonValueMapper(SYSTEM, SNAKE_CASE);
+    private final IonValueMapper ION_VALUE_MAPPER = new IonValueMapper(SYSTEM, SNAKE_CASE);
```
cowtowncoder: Why change from `static` to per-instance? Did it fail otherwise? (possibly due to null-value caching)

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
```diff
@@ -92,23 +100,49 @@ public void shouldBeAbleToDeserializeIncludingNullList() throws Exception {
...
 
     @Test
-    public void shouldBeAbleToDeserializeNullList() throws Exception {
-        IonValue ion = ion("{c:null.list}");
-
-        IonValueData data = ION_VALUE_MAPPER.readValue(ion, IonValueData.class);
+    public void shouldBeAbleToDeserializeNullToIonNull() throws Exception {
+        String ion = "{c:null}";
+        verifyNullDeserialization(ion, SYSTEM.newNull());
+        ION_VALUE_MAPPER.disable(IonParser.Feature.READ_NULL_AS_IONVALUE);
```
cowtowncoder: Should not try to change configuration of ION_VALUE_MAPPER after use, across methods; not guaranteed to take effect. Instead can create `ObjectReader`, call `ObjectReader.with()` and `.without()` to enable/disable features.

(also makes it easier to merge tests to 3.x as 3.x does not have methods to directly configure `ObjectMapper`)

cowtowncoder (commented):
Looks pretty good, couple of things to change.

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
seadbrane: > Should not try to change configuration of ION_VALUE_MAPPER after use,across methods; not guaranteed to take effect.

Why is that - caching?  

Using `ObjectReader` is not a great option here, since it doesn't support Ion natively.    Will create another `IonValueMapper`.

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
seadbrane: This was changed so that it can be configured in each test, but not needed if I create a new `IonValueMapper` for the `READ_NULL_AS_IONVALUE` cases.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
seadbrane: A couple of tests fail if this `_updateToken(JsonToken.VALUE_EMBEDDED_OBJECT);` is removed.  Perhaps a 3.0 cleanup activity.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
seadbrane: This is already there on line 125.

ion/src/test/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializerTest.java
cowtowncoder: Yes, indirectly caching of (de)serializers. Some of configuration settings affect construction and so effects remain if settings are changed.
Settings changed here are probably ok, although code won't work in 3.0 as mappers are truly immutable (builders only)

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/ionvalue/IonValueDeserializer.java
cowtowncoder: Ok thanks somehow missed it.

ion/src/main/java/com/fasterxml/jackson/dataformat/ion/IonParser.java
cowtowncoder: Ah. Makes sense there is a reason -- should have caught when these were added (or added comments to explain why/how)

Issue #431, opened 2023-12-26
Negative `BigInteger` values not encoded/decoded correctly


bgruber:
Hi!
I was doing some work with the CBOR part of the library, and I found that it didn't decode or encode negative BigIntegers as I understand the spec. For example:

```java
var negOne = BigInteger.ONE.negate();
var cborMapper = new ObjectMapper(new CBORFactory());
var bytes = cborMapper.writeValueAsBytes(negOne);
System.out.println(org.apache.commons.codec.binary.Hex.encode(bytes));
```
This prints out `c34101`, but [per this CBOR playground](https://cbor.me/?bytes=c34101), that's CBOR for a bigint of -2, not -1. 

Similarly on the read side:
```java
    var cborMapper = new ObjectMapper(new CBORFactory());
    var bytes =
        new byte[] {
          (byte) 0xC3,
          (byte) 0x50,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF,
          (byte) 0xFF
        };
    var read = cborMapper.readValue(bytes, BigInteger.class);
    System.out.println(read);
```
Prints `1`, instead of `-340282366920938463463374607431768211456` (-2^128). [CBOR playground](https://cbor.me/?bytes=c350ffffffffffffffffffffffffffffffff)

I think the issue is that the library is constructing the BigIntegers from the byte string, and then negating it, instead of taking the complement of the byte string (or "`-1 - n`" as stated by [the RFC](https://www.rfc-editor.org/rfc/rfc8949.html#name-bignums:~:text=For%20tag%20number%203%2C%20the%20value%20of%20the%20bignum%20is%20%2D1%20%2D%20n)) and constructing the BigInteger from that.

Thanks for all your work on jackson!

cowtowncoder:
That sounds plausible to me. Thank you for reporting this @bgruber.

But aside from confirming the case and proposing a fix, the tricky part is compatibility.
It is unfortunate that generation produces wrong value (essentially corrupt encoding); if this was only for decoding there'd be bit less problem.
[...]

PR #578, opened 2025-04-26
[CBOR] - Implement RFC compliant BigInteger bytes encoding & decoding 

iifawzi:
Hello, I discovered this issue #431 while working on the async parser for CBOR. 
This PR adds feature flags to control compliant encoding & decoding behavior as described in the RFC https://datatracker.ietf.org/doc/html/rfc8949#section-3.4.3

- For encoding: Changed calculation to `-1 - n` for negative values
- For decoding: 
  - Applied `-1 - n` formula
  - Updated to treat bytes as signed using `new BigInteger(1, _binaryValue)`

Changes are controlled by feature flags:
- `CBORGenerator.Feature.CORRECT_CBOR_NEGATIVE_BIGINT_ENCODING`
- `CBORParser.Feature.CORRECT_CBOR_NEGATIVE_BIGINT_DECODING`

Both flags default to false for backward compatibility.

Since changes are feature-flag controlled, I think they should be safe for 2.19.1 and 2.20 releases (I'm not sure about versioning after the release) @cowtowncoder I'll adjust base branch/comments based on your feedback.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
@@ -128,7 +128,7 @@ public class CBORGenerator extends GeneratorBase
          *
          * The default value is false for backwards compatibility.
          *
-         * @since 2.19.1
+         * @since 2.20.0
          */
 
         CORRECT_CBOR_NEGATIVE_BIGINT_ENCODING(false)
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
@@ -119,19 +119,16 @@ public class CBORGenerator extends GeneratorBase
         /**
          * Feature that determines how binary tagged negative BigInteger values are
          * encoded.
-         *
-         * When enabled, Ensures proper encoding of negative values
-         * (e.g., -1 is encoded [0xC3, 0x41, 0x00])
-         *
-         * When disabled, Maintains backwards compatibility with existing implementations
-         * (e.g., -1 is encoded [0xC3, 0x41, 0x01])
-         *
-         * The default value is false for backwards compatibility.
+         * When enabled, ensures proper encoding of negative values
+         * (e.g., -1 is encoded {@code [0xC3, 0x41, 0x00])}
+         * When disabled, maintains backwards compatibility with existing implementations
+         * (e.g., -1 is encoded {@code [0xC3, 0x41, 0x01])}
+         *<p>
+         * Default value is {@code false} for backwards-compatibility.
          *
          * @since 2.20.0
          */
-
-        CORRECT_CBOR_NEGATIVE_BIGINT_ENCODING(false)
+        ENCODE_USING_STANDARD_NEGATIVE_BIGINT_ENCODING(false)
         ;
 
         protected final boolean _defaultState;
@@ -1227,14 +1224,13 @@ public class CBORGenerator extends GeneratorBase
     // Main write method isolated so that it can be called directly
     // in cases where that is needed (to encode BigDecimal)
     protected void _write(BigInteger v) throws IOException {
-        /*
-         * Supported by using type tags, as per spec: major type for tag '6'; 5
+        /* Supported by using type tags, as per spec: major type for tag '6'; 5
          * LSB either 2 for positive bignum or 3 for negative bignum. And then
          * byte sequence that encode variable length integer.
          */
         if (v.signum() < 0) {
             _writeByte(BYTE_TAG_BIGNUM_NEG);
-            if (isEnabled(CBORGenerator.Feature.CORRECT_CBOR_NEGATIVE_BIGINT_ENCODING)) {
+            if (isEnabled(CBORGenerator.Feature.ENCODE_USING_STANDARD_NEGATIVE_BIGINT_ENCODING)) {
                 v = BigInteger.ONE.negate().subtract(v);
             } else {
                 v = v.negate();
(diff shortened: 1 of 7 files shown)


cowtowncoder:
First of all: thank you for working on this!

Second of all: Rats! With SemVer, we can't really merge that in 2.19 in a patch as that changes API.
So needs to go in 2.20 -- so 2.x branch is correct.

iifawzi:
> First of all: thank you for working on this!
> 
> Second of all: Rats! With SemVer, we can't really merge that in 2.19 in a patch as that changes API. So needs to go in 2.20 -- so 2.x branch is correct.

Thank you. so I understand even if it's managed by feature flags, we consider it an API change. 

I will update the comments, and will also update the tests to a more realistic case, as I realized even if it's possible to have BigInteger(-1), it's not really a big integer, and might be confusing given checking any online encoder of CBOR won't use big integer tag for -1 by default. I will include another test with an actual big integer for clarity. 

edit: marking it as a draft temporarily until I do more verifications and add more tests.

iifawzi:
The only point to comment on is that we're not following the preferred serialization as described https://www.rfc-editor.org/rfc/rfc8949.html#name-bignums, so encoding `-340282366920938463463374607431768211456` using jackson results in: 

```JAVA
byte[] expectedBytes = {
                (byte) 0xC3,
                (byte) 0x51, // 17 bytes - leading zero, 16 for the number
                (byte) 0x00, // LEADING Zero
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF,
                (byte) 0xFF
        };
```

which's still considered fine encoding as per the RFC as long as we're able to decode it `Decoders that understand these tags MUST be able to decode bignums that do have leading zeroes`.

Tests added in the mapper to verify we're able to decode both with/without leading zeros to the same correct value, while for backward compatibility, it remains the same, it returns different incorrect values with/without leading zeros. 


EDIT: we're fine with not following the preferred way anyway, it wasn't mentioned in the initial RFC (https://www.rfc-editor.org/rfc/rfc7049#section-2.4.2), only mentioned point is that decoder should be able to decode with/without leading zeros, which's what has been achieved through this PR.

cowtowncoder:
> Thank you. so I understand even if it's managed by feature flags, we consider it an API change.

More specifically: Addition of said Feature flags is an API change (functionality addition). Something to do in a minor release, but not in patch.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -1123,11 +1150,16 @@ protected JsonToken _handleTaggedBinary(TagList tags) throws IOException
...
-            BigInteger nr = new BigInteger(_binaryValue);
-            if (neg) {
-                nr = nr.negate();
-            }
-            _numberBigInt = nr;
+
+                BigInteger nr = new BigInteger(_binaryValue);
+                if (neg) {
+                    if (Feature.CORRECT_CBOR_NEGATIVE_BIGINT_DECODING.enabledIn(_formatFeatures)) {
+                        nr = BigInteger.ONE.negate().subtract(new BigInteger(1, _binaryValue));
```
cowtowncoder: This throws away `nr` constructed earlier... I think I'll refactor this to remove unnecessary calculation

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -115,6 +115,23 @@ public enum Feature implements FormatFeature {
...
+         *
+         * When disabled, Maintains backwards compatibility with existing implementations
+         * (e.g., -1 is encoded [0xC3, 0x41, 0x01])
+         *
+         * The default value is false for backwards compatibility.
+         *
+         * @since 2.20.0
+         */
+
+        CORRECT_CBOR_NEGATIVE_BIGINT_ENCODING(false)
```
cowtowncoder: I think I'll rename this to something like ENCODE_USING_STANDARD_BIGINT_ENCODING.

cowtowncoder:
Will make some minor changes wrt naming, then merge. Thank you.

One follow-up question: should we consider changing defaults for Jackson 3.0.0? I have mixed feelings about that -- on one hand, we would ideally use correct standard encoding/decoding. But then again this probably breaks handling by users unaware of changes.
But not changing it will perpetuate incorrect handling.

cowtowncoder (approved):
LGTM!

cowtowncoder:
Big thank you @iifawzi for solving this! I merged it in 2.20 & forward to 3.x.
We can consider changing defaults as a follow-up step.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
iifawzi: makes sense, thank you

iifawzi:
> Will make some minor changes wrt naming, then merge. Thank you.
> 
> One follow-up question: should we consider changing defaults for Jackson 3.0.0? I have mixed feelings about that -- on one hand, we would ideally use correct standard encoding/decoding. But then again this probably breaks handling by users unaware of changes. But not changing it will perpetuate incorrect handling.

I've been thinking about while working on the async parser, on whether it should have same experience or not. 

In my opinion, we should change defaults for 3.0.0, major releases meant to break compatibility (if needed), and I think in that case it's needed as current encoding leads to entirely incorrect values. However, given the consistency we had for encoding/decoding, users using Jackson for roundtrip wouldn't even notice the fix (unless they have plain assertions in binary). it will break only for users with partial usage of jackson (encoding using jackson and plain binary for decoding, or vise-versa), so I think better to default to true for 3.0.0.

If you agree, I can create a separate PR targeting 3.x

cowtowncoder:
@iifawzi Yes, I concur. PR for change would be welcome. I will create Issue to match PR with.

**EDIT**: #582

PR #588, opened 2025-05-05
Allow exposing CBOR undefined value as JsonToken.VALUE_EMBEDDED_OBJECT

### LINKED ISSUE #137: Allow exposing CBOR "undefined" value as `JsonToken.VALUE_EMBEDDED_OBJECT` (with embedded value of `null`)
(note: continuation from #93)

Although coercing "undefined" into `null` token works to some degree, it seems better to allow exposing it in a way that allows caller to distinguish between real `null` and "undefined".
Since we have to do this via Jackson API, tokens, the most straight-forward reliable mean is to either:

1. Return `JsonToken.VALUE_EMBEDDED_OBJECT`, with "value" of `null`
2. Add a new method in `CBORParser` to let caller determine if last token returned was for "undefined" (CBOR encoded value of `0xF7`)

Or, better yet, allow both.

For backwards compatibility (since 2.9.6, "undefined" has been returned as `JsonToken.VALUE_NULL`), we need a new `CBORParser.Feature` to enable (1).
New method (name To Be Determined), however, will be available in either case, regardless of token chosen to represent it.

iifawzi:
Hello, this PR resolves https://github.com/FasterXML/jackson-dataformats-binary/issues/137 by 
- Adding a feature flag to decode `undefined` value as `JsonToken.VALUE_EMBEDDED_OBJECT`
- Adding `isUndefined` method that could be used for both legacy handling and with embedded objects, as an indicator if the last token parsed is undefined or not.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORConstants.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORConstants.java
@@ -113,6 +113,11 @@ public final class CBORConstants
 
     public final static int INT_BREAK = 0xFF;
 
+    /**
+     * Marker for "undefined" value in CBOR spec.
+     *
+     * @since 2.20
+     */
     public final static int SIMPLE_VALUE_UNDEFINED = 0xF7;
 
     /*
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORConstants.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORConstants.java
@@ -113,6 +113,8 @@ public final class CBORConstants
 
     public final static int INT_BREAK = 0xFF;
 
+    public final static int SIMPLE_VALUE_UNDEFINED = 0xF7;
+
     /*
     /**********************************************************
     /* Basic UTF-8 decode/encode table
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -1939,10 +1939,15 @@ public class CBORParser extends ParserMinimalBase
      * and the token is reported as {@link JsonToken#VALUE_NULL}.
      *
      * @return {@code true} if current token is an `undefined`, {@code false} otherwise
+     *
      * @since 2.20
      */
     public boolean isUndefined() {
-        return (_inputBuffer[_inputPtr - 1] & 0xFF) == SIMPLE_VALUE_UNDEFINED;
+        if ((_currToken == JsonToken.VALUE_NULL) || (_currToken == JsonToken.VALUE_NULL)) {
+            return (_inputBuffer != null)
+                    && (_inputBuffer[_inputPtr - 1] & 0xFF) == SIMPLE_VALUE_UNDEFINED;
+        }
+        return false;
     }
 
     /*
@@ -3686,9 +3691,8 @@ expType, type, ch));
      * Helper method to encapsulate details of handling of mysterious `undefined` value
      * that is allowed to be used as something encoder could not handle (as per spec),
      * whatever the heck that should be.
-     *
      * <p>
-     * For backward compatibility with Jackson 2.9.6 to 2.19, this value is decoded
+     * For backward compatibility with Jackson 2.10 to 2.19, this value is decoded
      * as {@link JsonToken#VALUE_NULL} by default.
      * <p>
      *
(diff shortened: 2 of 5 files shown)


iifawzi:
Defaults for 3.0 should be true, I believe, so it could take place in a separate PR

iifawzi:
I'm not sure if we need to touch `getEmbeddedObject`, now it will return null, I couldn't think of any other alternative than returning the string "undefined" but not feeling so strong about it. I think null maybe remains valid..

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORConstants.java
```diff
@@ -113,6 +113,8 @@ public final class CBORConstants
 
     public final static int INT_BREAK = 0xFF;
 
+    public final static int SIMPLE_VALUE_UNDEFINED = 0xF7;
```
cowtowncoder, line 121: Could use a Javadoc

cowtowncoder:
> I'm not sure if we need to touch `getEmbeddedObject`, now it will return null, I couldn't think of any other alternative than returning the string "undefined" but not feeling so strong about it. I think null maybe remains valid..

Good question. I don't think constant String would be any better.

But somehow not setting anything does seem odd. I think we can go with that; as long as behavior is tested I think we are ok. Will add one specific suggestion on code, for clearing `_binaryValue` explicitly: while redundant, works as sort of documentation.

The only (?) concern I have is about future embedded values we might want to return, and whether `null` having specific semantics might be problematic. But that's hypothetical for now so... no need to solve problem we do not have.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -1915,6 +1931,20 @@ private final byte[] _getBinaryFromString(Base64Variant variant) throws IOExcept
...
+     * <p>
+     * This method allows distinguishing between real {@code null} and `undefined`,
+     * even if {@link CBORParser.Feature#HANDLE_UNDEFINED_AS_EMBEDDED_OBJECT} is disabled
+     * and the token is reported as {@link JsonToken#VALUE_NULL}.
+     *
+     * @return {@code true} if current token is an `undefined`, {@code false} otherwise
+     * @since 2.20
+     */
+    public boolean isUndefined() {
+        return (_inputBuffer[_inputPtr - 1] & 0xFF) == SIMPLE_VALUE_UNDEFINED;
```
cowtowncoder: Hmmmh. Not a fan of this approach but... I guess it'll do. Let's preface it by

    (_currToken == JsonToken.VALUE_EMBEDDED_OBJECT || _currToken == JsonToken.VALUE_NULL)

to avoid likelihood of incorrect deterimation (binary segment ending with byte `0xF7`).

Also: not 100% sure if bounds checks would be needed.
Actually, must check that `_inputBuffer` is not `null` -- `_releaseBuffers()` clears it.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -3656,13 +3686,22 @@ private final static long _long(int i1, int i2)
...
+     *
+     * since 2.20 If {@link CBORParser.Feature#HANDLE_UNDEFINED_AS_EMBEDDED_OBJECT} is enabled,
+     * the value will instead be decoded as {@link JsonToken#VALUE_EMBEDDED_OBJECT}
+     * with an embedded value of {@code null}.
      *
      * @since 2.9.6
      */
-    protected JsonToken _decodeUndefinedValue() throws IOException {
+    protected JsonToken _decodeUndefinedValue() {
+        if (CBORParser.Feature.HANDLE_UNDEFINED_AS_EMBEDDED_OBJECT.enabledIn(_formatFeatures)) {
```
cowtowncoder: Let's add `_binaryValue = null;` just as sort of documentation of intent. I know it should be cleared by `nextToken()` and other places but it seems better be explicit in this particular case.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder: Actually I can add that.

cowtowncoder:
Ok I added release notes and made minor tweaks -- actually can add Javadoc too and then merge.

cowtowncoder (approved):

cowtowncoder:
@iifawzi Merged into 2.20, will merge forward into 3.x, so separate PR just needed for defaults change and unit tests (likely).

iifawzi:
> > I'm not sure if we need to touch `getEmbeddedObject`, now it will return null, I couldn't think of any other alternative than returning the string "undefined" but not feeling so strong about it. I think null maybe remains valid..
> 
> Good question. I don't think constant String would be any better.
> 
> But somehow not setting anything does seem odd. I think we can go with that; as long as behavior is tested I think we are ok. Will add one specific suggestion on code, for clearing `_binaryValue` explicitly: while redundant, works as sort of documentation.
> 
> The only (?) concern I have is about future embedded values we might want to return, and whether `null` having specific semantics might be problematic. But that's hypothetical for now so... no need to solve problem we do not have.

Thanks for doing the changes, I also believe for future embedded values (SimpleValue) we need to return the value itself as null would be meaningless and unexpected, but yes, maybe we discuss it in the corresponding pr later

Thank you for completing the missing points

Issue #587, opened 2025-05-04
Allow exposing CBOR Simple values as `JsonToken.VALUE_EMBEDDED_OBJECT` with a feature flag


iifawzi:
Hello, currently, we're decoding Simple values from [Major type (7) ](https://www.rfc-editor.org/rfc/rfc7049#section-2.3) as `VALUE_NUMBER_INT` Token, where it's better to be decoded using a special type (I agree and assume VALUE_EMBEDDED_OBJECT would work and no need for additional type as commented) 

https://github.com/FasterXML/jackson-dataformats-binary/blob/63fdc54928485932a881029f115e934f3546a9b6/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L3679-L3705

I'm writing this issue so we track it, where we can introduce that change with a feature flag default to false for earlier versions, and true for 3.0 onward.

iifawzi:
Hey @cowtowncoder, not sure if you would like different behavior on how should we move with handling these values, i'm creating this initial issue to track it and to be able to create a PR against it

cowtowncoder:
Ok so at high level that sounds good, wrt change to `JsonToken.VALUE_EMBEDDED_OBJECT`, opt-in for 2.x, default for 3.0.

One immediate idea is that although there is `JsonToken.VALUE_EMBEDDED_OBJECT` question then would be which Object type to expose it as. I think it'd make sense to specify new simple wrapper type and not use `Integer` (or `Long`, if need be).

I don't have strong opinion on naming but looks like term "tag" is used, so maybe `SemanticTag`.
Doing this would make things bit more type-safe, without adding much overhead.

**EDIT**: realized these are NOT Tags but specifically "Simple values". So something reflecting that, like `SimpleValue`, would probably make sense.

iifawzi:
> Ok so at high level that sounds good, wrt change to `JsonToken.VALUE_EMBEDDED_OBJECT`, opt-in for 2.x, default for 3.0.
> 
> One immediate idea is that although there is `JsonToken.VALUE_EMBEDDED_OBJECT` question then would be which Object type to expose it as. I think it'd make sense to specify new simple wrapper type and not use `Integer` (or `Long`, if need be).
> 
> I don't have strong opinion on naming but looks like term "tag" is used, so maybe `SemanticTag`. Doing this would make things bit more type-safe, without adding much overhead.
> 
[...]

PR #590, opened 2025-05-10
Allow exposing CBOR simple values as VALUE_EMBEDDED_OBJECT

iifawzi:
Hello, this PR is resolving #587 to allow exposing simple values as value embedded objects

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -1865,7 +1865,7 @@ public class CBORParser extends ParserMinimalBase
         if (_tokenIncomplete) {
             _finishToken();
         }
-        if (_currToken == JsonToken.VALUE_EMBEDDED_OBJECT ) {
+        if (_currToken == JsonToken.VALUE_EMBEDDED_OBJECT) {
             if (_simpleValue != null) {
                 return _simpleValue;
             }
@@ -1958,11 +1958,11 @@ public class CBORParser extends ParserMinimalBase
     /**
      * Checking whether the current token represents an `undefined` value (0xF7).
      * <p>
-     * This method allows distinguishing between real {@code null} and `undefined`,
+     * This method allows distinguishing between real {@code null} and {@code undefined},
      * even if {@link CBORParser.Feature#READ_UNDEFINED_AS_EMBEDDED_OBJECT} is disabled
      * and the token is reported as {@link JsonToken#VALUE_NULL}.
      *
-     * @return {@code true} if current token is an `undefined`, {@code false} otherwise
+     * @return {@code true} if current token is an {@code undefined}, {@code false} otherwise
      *
      * @since 2.20
      */
@@ -3739,7 +3739,7 @@ expType, type, ch));
      * and exposing them as expected token.
      * <p>
      * Starting with Jackson 2.20, this behavior can be changed by enabling the
-     * {@link com.fasterxml.jackson.dataformat.cbor.CBORParser.Feature#READ_SIMPLE_VALUE_AS_EMBEDDED_OBJECT}
+     * {@link CBORParser.Feature#READ_SIMPLE_VALUE_AS_EMBEDDED_OBJECT}
      * feature, in which case simple values are returned as {@link JsonToken#VALUE_EMBEDDED_OBJECT} with an
      * embedded {@link CBORSimpleValue} instance.
      *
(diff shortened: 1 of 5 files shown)


iifawzi:
A separate PR is needed to change defaults for 3.0. Most probably after the merge of this PR I will create that PR with defaults changes for this one, and other two features added recently

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -59,7 +59,21 @@ public enum Feature implements FormatFeature
          *
          * @since 2.20
          */
-        HANDLE_UNDEFINED_AS_EMBEDDED_OBJECT(false)
+        HANDLE_UNDEFINED_AS_EMBEDDED_OBJECT(false),
```
cowtowncoder: Just realized that I'll probably want change naming here slightly; instead of "handle", either "expose" or maybe "read" or "return". Will go ahead and make that change for both new Feature names.

I think I'll call with `READ_` since that is most commonly used by other Jackson components.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -823,9 +841,9 @@ public JsonToken nextToken() throws IOException
             _skipIncomplete();
         }
         _tokenInputTotal = _currInputProcessed + _inputPtr;
-        // also: clear any data retained so far
-        _numTypesValid = NR_UNKNOWN;
-        _binaryValue = null;
+
+        // also: clear any data retained so far.
+        clearRetainedValues();
```
cowtowncoder, line 851: There are a few other places this must be done (everywhere where `_binaryValue = null;` is done)

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
iifawzi: I think it could be better; I'm not good at naming these feature flags, though, so thank you! :-D

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
iifawzi: I thought clearing it once at the beginning of nextToken could be enough, given it's used only once for major 7, and there is no partial assignment (always one byte).

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder: np, that's what I am ultimately responsible for :)
(not that it was bad name per se, just trying to keep uniform terminology etc)

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder: I wouldn't want to count on state not getting cleared for `nextFieldName()` variants, as they won't call `nextToken()`.
It may be that a combination of things could keep things consistent (as in, not exposing stale simple value) but when adding new features invariants might not hold.

But if you are confident after looking, that's fine, just LMK.

iifawzi: on the side; so far I've been exploring the async impl of CBOR; and I'm almost done with my draft, building almost all major types to explore the spec. In next weeks, I will start work on the actual organized PR, However, I'd like to know if you think there are higher priority issues that the community would like to be fixed. I'm enjoying contributing to the project in my free time, so aligning with priorities will make it even more valuable. 

After the async parser, I believe looking at Avro issues is worth it, as I think it's the most used (just a guess from the number of issues)

Would appreciate your thoughts.

iifawzi: > I wouldn't want to count on state not getting cleared for `nextFieldName()` variants, as they won't call `nextToken()`. It may be that a combination of things could keep things consistent (as in, not exposing stale simple value) but when adding new features invariants might not hold.
> 
> But if you are confident after looking, that's fine, just LMK.

Aha, I got it, and well, I'm not 100% confident about the different combinations that might happen that can lead to exposing stale values, so yeah, better not to count on that. I will commit the change to clear it in places where binary values is cleared

cowtowncoder: @iifawzi On async impl: that sounds like something that would be well received so +1 for that. Avro does seem to get more attention that I'd have expected so that's a good thing to follow up on too. And maybe logical type support and whatever "new" (... I haven't been following up spec development) features for CBOR would be good as well.
(Avro, too, has logical type concept).

And ideally Proto3 would be supported by protobuf module. There's also question of replacing protoc-parser (reader for schema) that is deprecated with library that is to replace it (by same authors). I haven't had time to follow up on this, I think there's an issue filed to request it.

iifawzi: Thanks a lot for the insights @cowtowncoder! I’ll keep those points in mind as I move forward after the async impl

cowtowncoder (approved):
LGTM!

cowtowncoder:
Thank you again, @iifawzi. I am now merging to 3.x, after which you can do PR to change defaults for features (which are now `CBORReadFeature`s in 3.0).

Issue #589, opened 2025-05-07
AvroSchema: Does not include base class for records with subclasses


raphw:
I think I encountered an issue in the schema generation of records that are unions: https://github.com/FasterXML/jackson-dataformats-binary/blob/605e79ff2236950a0f0548681717dc177ba65a9d/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java#L53

Given the following example (using Jakarta annotations and the introspector that follows with it):

```java
@XmlType(name = "Super")
@XmlAccessType(FIELD)
@XmlSeeAlso(Sub.class)
class Super { 
    @XmlElement(required = true)
    String foo; // getter, setter
}

@XmlType(name = "Sub")
@XmlAccessType(FIELD)
class Sub extends Super { 
    @XmlElement(required = true)
    String bar; // getter, setter
}
```

where the class `Sub` is correctly identified, a value of `Super` is not allowed in the created union. 

To avoid this issue, `unionSchemas` should also include `type`, if `type` is non-abstract.

cowtowncoder:
I think a slightly more extensive example would be useful to show how types in question are used (which specific class is schema being generated etc).
For example, `Super` here has nothing to link it to `Sub` (`@JsonSubTypes` would solve that), but not sure that is what is being problem here.

raphw:
In my case, I am using Jakarta annotations and the respective introspector (see edit above). But any introspector would do. The problem is that the introspector is requested to return the subclasses and does so correctly: https://github.com/FasterXML/jackson-dataformats-binary/blob/605e79ff2236950a0f0548681717dc177ba65a9d/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java#L45

The base class is no longer considered, what is correct if it is abstract. However, a non-abstract class should be considered.

cowtowncoder:
Ah. Good to know the issue.

@MichalFoksa WDYT? Sounds like it could be an easy fix if you agree this is something to fix.

MichalFoksa:
Hi there,
I could reproduce the problem and I have ugly fix. 
Let me refine it somehow.

cowtowncoder:
[...]

PR #593, opened 2025-05-14
[AVRO] #589: Fix schema not including base class for records with subclasses

MichalFoksa:
Supported cases:

### 1. Concrete class annotated with @JsonSubTypes
When a concrete (non-abstract) class is annotated with @JsonSubTypes, then Avro type of the annotated class is part of resulting union.

```java
 @JsonSubTypes({
        @JsonSubTypes.Type(value = Apple.class),
        @JsonSubTypes.Type(value = Pear.class),
 })
class Fruit { }

class Apple extends Fruit {}
class Pear extends Fruit {}
```

Fruit schema:
```avroschema
[
  {name: Fruit, type: record, fields: [...]},
  {name: Apple, type: record, fields: [...]},
  {name: Orange, type: record, fields: [...]}
]
```
Annotated class, Fruit, is part of the schema:

This was not supported case, annotated class was missing in union.

#### 2. Abstract class annotated with @JsonSubTypes
When an abstract class is annotated with @JsonSubTypes, then abstract class is not in union.

```java
 @JsonSubTypes({
        @JsonSubTypes.Type(value = Apple.class),
        @JsonSubTypes.Type(value = Pear.class),
 })
class AbstractFruit { }          // <----  abstract class

class Apple extends Fruit {}
class Pear extends Fruit {}
```

AbstractFruit schema:
```avroschema 
[
  {name: Apple, type: record, fields: [...]},
  {name: Orange, type: record, fields: [...]}
]
```
Annotated abstract base class is not part of the schema:

This was supported case.

#### 3. Abstract class somewhere in the middle of @JsonSubTypes hierarchy
When an abstract class is somewhere along @JsonSubTypes hierarchy, it does not end up in union.

```java
@JsonSubTypes({
        @JsonSubTypes.Type(value = AbstractWaterVehicle.class),
})
class Vehicle {
}

@JsonSubTypes({
        @JsonSubTypes.Type(value = Boat.class),
})
abstract class AbstractWaterVehicle extends Vehicle { }   // <---- abstract class
class Boat extends AbstractWaterVehicle { }
```

Vehicle schema:
```avroschema
[
  {name: Vehicle, type: record, fields: [...]},
  {name: Boat, type: record, fields: [...]},
]
```
AbstractWaterVehicle class is not in union.

This was not supported case, AbstractWaterVehicle would be part of union

### 4. Class referenced twice in @JsonSubTypes hierarchy
When a class is referenced twice in @JsonSubTypes hierarchy, it occurs only once in union.

```java
// Helium is twice in subtypes hierarchy, once as `ElementInterface` subtype and second time as subtype of `AbstractGas` subtype.
@JsonSubTypes({
        @JsonSubTypes.Type(value = Gas.class),
        @JsonSubTypes.Type(value = Helium.class),       // <---- 1. Helium 
})
interface ElementInterface { }

@JsonSubTypes({
        @JsonSubTypes.Type(value = Helium.class),       // <---- 2. Helium
        @JsonSubTypes.Type(value = Oxygen.class),
})
static class Gas implements ElementInterface { }

class Helium extends Gas { }
class Oxygen extends Gas { }
```

ElementInterface schema:
```avroschema
[
  {name: Gas, type: record, fields: [...]},
  {name: Helium, type: record, fields: [...]},
  {name: Oxygen, type: record, fields: [...]}
]
```

`Helium`  class is not duplicated in union.

### 5. Base class explicitly in @JsonSubTypes or @Union annotations
When a class is subclass of itself.

```java
 @JsonSubTypes({
        @JsonSubTypes.Type(value = Fruit.class),   // <----  Fruit is annotated class and also in @JsonSubTypes
        @JsonSubTypes.Type(value = Apple.class),
        @JsonSubTypes.Type(value = Pear.class),
 })
class Fruit { } // or interface

class Apple extends Fruit {}
class Pear extends Fruit {}
```

```java
 @Uion({
        Fruit.class,                              // <----  Fruit is annotated class and also in @Union
        Apple.class, Pear.class,
 })
class Fruit { }  // or interface

class Apple extends Fruit {}
class Pear extends Fruit {}
```

Both cases above would lead to endless loop and StackOverflowError. 

---
See `PolymorphicTypeAnnotationsTest.class`.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
@@ -85,25 +85,56 @@ public class RecordVisitor
 
             List<NamedType> subTypes = getProvider().getAnnotationIntrospector().findSubtypes(bean.getClassInfo());
             if (subTypes != null && !subTypes.isEmpty()) {
-                List<Schema> unionSchemas = new ArrayList<>();
-                // Initialize with this schema
-                if (_type.isConcrete()) {
-                    unionSchemas.add(_typeSchema);
-                }
+
+                // At this point calculating hashCode for _typeSchema fails with NPE because RecordSchema.fields is NULL
+                // (see org.apache.avro.Schema.RecordSchema#computeHash).
+                // Therefore, _typeSchema must be added into union at very end, or unionSchemas must not be HashSet (or any
+                // other type calling hashCode() for equality check).
+                Set<Schema> unionSchemas = new HashSet<>();
+                // ArrayList<Schema> unionSchemas = new ArrayList<>();
+
+                // IdentityHashMap is used because it is using reference-equality.
+                // Set<Schema> unionSchemas = Collections.newSetFromMap(new IdentityHashMap<>());
+
+                // Initialize with this schema is
+                // if (_type.isConcrete()) {
+                //   unionSchemas.add(_typeSchema);
+                // }
+
                 try {
                     for (NamedType subType : subTypes) {
                         JsonSerializer<?> ser = getProvider().findValueSerializer(subType.getType());
                         VisitorFormatWrapperImpl visitor = _visitorWrapper.createChildWrapper();
                         ser.acceptJsonFormatVisitor(visitor, getProvider().getTypeFactory().constructType(subType.getType()));
                         Schema subTypeSchema = visitor.getAvroSchema();
-                        // If subType schema is also UNION, include all its types into this union
+                        // Add subType schema into this union, unless it is already there.
+                        // When subType schema is itself a union, include all its types into this union
                         if (subTypeSchema.getType() == Type.UNION) {
+//                            subTypeSchema.getTypes().stream()
+//                                    .filter(unionElement -> !unionSchemas.contains(unionElement))
+//                                    .forEach(unionSchemas::add);
+                            // or
+//                            for( Schema unionElement: subTypeSchema.getTypes()) {
+//                                if (unionSchemas.contains(unionElement)) {
+//                                    continue;
+//                                }
+//                                unionSchemas.add(unionElement);
+//                            }
                             unionSchemas.addAll(subTypeSchema.getTypes());
                         } else {
+//                            if (!unionSchemas.contains(subTypeSchema)) {
+//                                unionSchemas.add(subTypeSchema);
+//                            }
                             unionSchemas.add(subTypeSchema);
                         }
                     }
-                    _avroSchema = Schema.createUnion(unionSchemas);
+
+                    ArrayList<Schema> unionList = new ArrayList<>(unionSchemas);
+                    // add _type schema into union
+                    if (_type.isConcrete()) {
+                        unionList.add(_typeSchema);
+                    }
+                    _avroSchema = Schema.createUnion(unionList);
                 } catch (JsonMappingException jme) {
                     throw new RuntimeException("Failed to build schema", jme);
                 }
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
@@ -85,17 +85,26 @@ public class RecordVisitor
 
             List<NamedType> subTypes = getProvider().getAnnotationIntrospector().findSubtypes(bean.getClassInfo());
             if (subTypes != null && !subTypes.isEmpty()) {
+                // alreadySeenClasses prevents subType processing in endless loop
+                Set<Class<?>> alreadySeenClasses = new HashSet<>();
+                alreadySeenClasses.add(_type.getRawClass());
+
                 // At this point calculating hashCode for _typeSchema fails with NPE because RecordSchema.fields is NULL
                 // see org.apache.avro.Schema.RecordSchema#computeHash.
                 // Therefore, unionSchemas must not be HashSet (or any other type using hashCode() for equality check).
+                // Set ensures that each subType schema is once in resulting union.
                 // IdentityHashMap is used because it is using reference-equality.
                 Set<Schema> unionSchemas = Collections.newSetFromMap(new IdentityHashMap<>());
                 // Initialize with this schema
                 if (_type.isConcrete()) {
                     unionSchemas.add(_typeSchema);
                 }
+
                 try {
                     for (NamedType subType : subTypes) {
+                        if (!alreadySeenClasses.add(subType.getType())) {
+                            continue;
+                        }
                         JsonSerializer<?> ser = getProvider().findValueSerializer(subType.getType());
                         VisitorFormatWrapperImpl visitor = _visitorWrapper.createChildWrapper();
                         ser.acceptJsonFormatVisitor(visitor, getProvider().getTypeFactory().constructType(subType.getType()));
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
@@ -26,15 +26,16 @@ public class RecordVisitor
     protected final VisitorFormatWrapperImpl _visitorWrapper;
 
     /**
-     * Tracks if the schema for this record has been overridden (by an annotation or other means), and calls to the {@code property} and
-     * {@code optionalProperty} methods should be ignored.
+     * Tracks if the schema for this record has been overridden (by an annotation or other means),
+     * and calls to the {@code property} and {@code optionalProperty} methods should be ignored.
      */
     protected final boolean _overridden;
 
     /**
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Sounds good. Was about to suggest that due to `RecordVisitor` changes (addition of a Field) should go against 2.x not 2.19 but since it's `private` field maybe it can be considered internal (non-API) change.

MichalFoksa:
I have problem to choose most obvious handling of duplicates in unionSchemas (inside `if (subTypes != null && !subTypes.isEmpty())` condition).

1.  when I use HashSet then `_typeSchema` must be added into union last.
2.  when I use ArrayList, then I have to call `unionSchemas .contains(subTypeSchema)` before adding subTypeSchema into `unionSchemas`. However  ArrayList.contains() is only object reference-equality check.

3.  If reference-equality check is enough for `unionSchemas` , then `IdentityHashMap` can be used to create a Set (`Set<Schema> unionSchemas = Collections.newSetFromMap(new IdentityHashMap<>())` ) and we have simple code.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
```diff
@@ -31,6 +31,33 @@ public class RecordVisitor
...
+     * When _type = Fruit.class
+     * Then
+     * _avroSchema if Fruit.class is union of Fruit record, Apple record and Orange record schemas: [
+     *     { name: Fruit, type: record, field: [..] }, <--- _typeSchema points here
+     *     { name: Apple, type: record, field: [..] },
+     *     { name: Orange, type: record, field: [..]}
+     *   ]
+     * _typeSchema points to Fruit.class without subtypes record schema
+     *
+     * FIXME: When _thisSchema is not null, then _overridden must be true, therefore (_overridden == true) can be replaced with (_thisSchema != null),
```
cowtowncoder: _thisSchema? Should it be _typeSchema? (or should _typeSchema be changed to _thisSchema)

cowtowncoder:
> I have problem to choose most obvious handling of duplicates in unionSchemas (inside `if (subTypes != null && !subTypes.isEmpty())` condition).
> 
>     1. when I use HashSet then `_typeSchema` must be added into union last.
> 
>     2. when I use ArrayList, then I have to call `unionSchemas .contains(subTypeSchema)` before adding subTypeSchema into `unionSchemas`. However  ArrayList.contains() is only object reference-equality check.
> 
>     3. If reference-equality check is enough for `unionSchemas` , then `IdentityHashMap` can be used to create a Set (`Set<Schema> unionSchemas = Collections.newSetFromMap(new IdentityHashMap<>())` ) and we have simple code.

Would it make sense to first use a `Map` with Java type being resolved (either `JavaType`, if available, or `Class` if not) as key; and only after collecting use "values" to create whatever is needed (if it must be `Set`, then identity; otherwise `List` if we can use `Collection`)?

I may be missing/misunderstanding something, so apologies if above makes no sense.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
MichalFoksa: Right, fixed it. 

The comment is from when  the variable `_typeSchema` was called `_thisSchema`.

cowtowncoder: Yeah so I figured.

MichalFoksa:
@cowtowncoder 

I am done. Have a look at it PLS.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
```diff
@@ -31,6 +31,33 @@ public class RecordVisitor
...
+     * Then
+     * _avroSchema if Fruit.class is union of Fruit record, Apple record and Orange record schemas: [
+     *     { name: Fruit, type: record, fields: [..] }, <--- _typeSchema points here
+     *     { name: Apple, type: record, fields: [..] },
+     *     { name: Orange, type: record, fields: [..]}
+     *   ]
+     * _typeSchema points to Fruit.class without subtypes record schema
+     *
+     * FIXME: When _typeSchema is not null, then _overridden must be true, therefore (_overridden == true) can be replaced with (_typeSchema != null),
+     * but it might be considered API change cause _overridden has protected access modifier.
```
cowtowncoder, line 59: We can do that as a follow-up for 2.20 (2.x branch)

cowtowncoder:
@MichalFoksa Thank you! Had a quick look and things look better. Will try to properly review later tonight, get it merged hopefully.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
```diff
@@ -42,32 +75,62 @@ public RecordVisitor(SerializerProvider p, JavaType type, VisitorFormatWrapperIm
...
+                alreadySeenClasses.add(_type.getRawClass());
+
+                // At this point calculating hashCode for _typeSchema fails with
+                // NPE because RecordSchema.fields is NULL
+                // (see org.apache.avro.Schema.RecordSchema#computeHash).
+                // Therefore, unionSchemas must not be HashSet (or any other type
+                // using hashCode() for equality check).
+                // Set ensures that each subType schema is once in resulting union.
+                // IdentityHashMap is used because it is using reference-equality.
+                final Set<Schema> unionSchemas = Collections.newSetFromMap(new IdentityHashMap<>());
```
cowtowncoder, line 106: But if we have identity-based `Set`, where the only match is pure identity, why not just use `ArrayList` instead of `Set` to `add`/`addAll` to? Or do we get literal same `Schema` instances somehow?

Making that chance will not fail any of added tests at least.

MichalFoksa: > Or do we get literal same Schema instances somehow?

Yes. It happens when a class is references multiple times in hierarchy (maybe by mistake).

In the example below Helium is subtype of Element and also subtype of Gas. 

Where Gas schema is union of 
  - Gas (itself), 
  - Helium 
  - Oxygen.

Element schema is union of all type in
  - Gas schema (Gas, Helium and Oxygen) 
  - Helium

When you collect all types togetjer without identity check you would get invalid union of 
  - Gas
  - Helium
  - Oxygen
  - Helium again 

because Helium is twice.


See use case 4 "Class referenced twice in @JsonSubTypes hierarchy**" above or `PolymorphicTypeAnnotationsTest.class_is_referenced_twice_in_hierarchy_test` test.

Change `unionSchemas` to ArrayList and the test will fail with   
> Failed to generate `AvroSchema` for `ElementInterface`, problem: Duplicate in union: `Helium`

```
 @JsonSubTypes({
   Type( Gas.class )
   Type( Helium.class ) })                         <---- First Helium occurrence 
 +---------------------+
 |      Element        |
 |    ( interface )    |
 +---------------------+
     ▲              ▲   @JsonSubTypes({
     |               |    Type( Helium.class )     <---- Second Helium occurrence
     |               |    Type( Oxygen.class ) })
     |     +---------------------+
     |     |        Gas          |
     |     +---------------------+
     |              ▲
     |              | 
     |     +--------+--------+
     |     |                 |
+-----------------+  +-----------------+
|     Helium      |  |     Oxygen      |
+-----------------+  +-----------------+
```

cowtowncoder: Ok yes, right, but: I _think_ duplicates are actually caught by `alreadySeenClasses` -- so I don't think `unionSchemas` prevent any duplicates. So that one could -- I think? -- be simple `ArrayList`

MichalFoksa: Both duplicate checks are needed, each check is on different level.

cowtowncoder: But like I said, changing Set to List did not fail any of the tests. So something is odd.
I think I'll try to do that again just to make sure I did not imagine it.

MichalFoksa: :( Hmmm, just tested:

`final ArrayList<Schema> unionSchemas = new ArrayList<>();`

Fails `PolymorphicTypeAnnotationsTest#class_is_referenced_twice_in_hierarchy_test` test.

cowtowncoder: Ok I must have been hallucinating. Apologies for noise.

cowtowncoder (approved):

cowtowncoder:
Merged all the way to 3.x, things working. Phew!

Next PR could be for minor changes for 2.20 (2.x), making fields final.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
MichalFoksa: no problem, it happens.

PR #603, opened 2025-07-09
Add support for writing unsigned numbers to CBOR

Radiokot:
Hey guys,

In a project I use Jackson CBOR dataformat for, the CBOR encoded in Kotlin is decoded by software expecting `uint64`. This PR implements support for writing unsigned numbers (long, int) in custom serializers, so values negative in Java do not end up negative in the CBOR. (https://datatracker.ietf.org/doc/html/rfc7049#appendix-A, https://datatracker.ietf.org/doc/html/rfc7049#appendix-B)

I've added generator methods and tests for them, but I am currently not sure what to do with the parser. In fact, an unsigned number can be parsed as BigInteger which is then converted to long via `BigInteger::toLong`, but it lacks the boundary check.

Currently, this can be used with custom serializers like this:
```kotlin
class UnsignedLongSerializer : JsonSerializer<Long>() {
    override fun serialize(value: Long, generator: JsonGenerator, serializerProvider: SerializerProvider) =
        (generator as CBORGenerator).writeNumberUnsigned(value)
}

class UnsignedLongDeserializer : JsonDeserializer<Long>() {
    override fun deserialize(parser: JsonParser, deserializationContext: DeserializationContext): Long =
        (parser as CBORParser).bigIntegerValue.toLong()
}

data class Account
@JsonCreator
constructor(
    val name: String,
    @JsonSerialize(using = UnsignedLongSerializer::class)
    @JsonDeserialize(using = UnsignedLongDeserializer::class)
    val balance: Long,
)
```

Do you think if support for unsigned numbers is needed in Jackson? If so, what should I add to proceed?

Thanks.

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
@@ -1142,6 +1142,20 @@ public class CBORGenerator extends GeneratorBase
         _writeByte(BYTE_NULL);
     }
 
+    /**
+     * Method for outputting given value as an unsigned integer.
+     * Can be called in any context where a value is expected
+     * (Array value, Object field value, root-level value).
+     *
+     * @param i Number value to write
+     * @throws IOException if there is either an underlying I/O problem or encoding
+     *                     issue at format layer
+     */
+    public void writeNumberUnsigned(int i) throws IOException {
+        _verifyValueWrite("write number unsigned");
+        _writeIntMinimal(PREFIX_TYPE_INT_POS, i);
+    }
+
     @Override
     public void writeNumber(int i) throws IOException {
         _verifyValueWrite("write number");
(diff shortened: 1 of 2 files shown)


cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/gen/GeneratorSimpleTest.java
```diff
@@ -1,17 +1,16 @@
 package com.fasterxml.jackson.dataformat.cbor.gen;
 
+import com.fasterxml.jackson.core.JsonGenerationException;
```
cowtowncoder: Please do not re-order imports: existing ordering is as per style guide.

cowtowncoder:
On reader side: not 100% sure. Was about to suggest caller has to deal with it but... haven't thought it through really.

As to support needed/useful; I think what you propose on writing makes sense as sort of "least we can do" approach.

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/gen/GeneratorSimpleTest.java
Radiokot: Got it, order restored.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -1142,6 +1142,20 @@ public void writeNull() throws IOException {
...
 
+    /**
+     * Method for outputting given value as an unsigned integer.
+     * Can be called in any context where a value is expected
+     * (Array value, Object field value, root-level value).
+     *
+     * @param i Number value to write
+     * @throws IOException if there is either an underlying I/O problem or encoding
+     *                     issue at format layer
+     */
```
cowtowncoder, line 1154: Please add

    @since 2.20

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORGenerator.java
```diff
@@ -1183,6 +1197,33 @@ public void writeNumber(int i) throws IOException {
...
 
+    /**
+     * Method for outputting given value as an unsigned integer.
+     * Can be called in any context where a value is expected
+     * (Array value, Object field value, root-level value).
+     *
+     * @param l Number value to write
+     * @throws IOException if there is either an underlying I/O problem or encoding
+     *                     issue at format layer
+     */
```
cowtowncoder, line 1210: Add

    @since 2.20 

here as well

cowtowncoder (approved):

cowtowncoder:
Ok, looks good to me.

Just one process thing before I merge this PR: unless we have gotten one already (I don't think so but if we have LMK), we'll need a CLA; doc is here:

it only needs to be sent once before the first contribution. The usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
Once we have it I can merge this PR & it's good for any other PRs for all Jackson repos.

Thank you again, looking forward to merging!

cowtowncoder:
CLA received, hoping to do final review & merge later tonight.

cowtowncoder (approved):

PR #613, opened 2025-09-18
Fixes #601: Jackson subtype schema unions are non-deterministic

### LINKED ISSUE #601: Jackson subtype Avro schema unions are non-deterministic and therefore incompatible with each other

beseder1:
(fixes: #601)

With the current implementation of the `RecordVisitor`, Avro schema unions are handled in a non-deterministic way. This can cause deserialization errors when different `AvroMapper` instances are used for serialization and deserialization.

Specifically, because the subtype ordering is unstable, deserialization may attempt to resolve a serialized object against the wrong subtype, resulting in errors.

This PR makes the ordering deterministic by using an `ArrayList` instead of a `Set`, ensuring consistent behavior across serialization and deserialization.

### DIFF
--- a/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
+++ b/avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
@@ -98,9 +98,8 @@ public class RecordVisitor
                 // (see org.apache.avro.Schema.RecordSchema#computeHash).
                 // Therefore, unionSchemas must not be HashSet (or any other type
                 // using hashCode() for equality check).
-                // Set ensures that each subType schema is once in resulting union.
-                // IdentityHashMap is used because it is using reference-equality.
-                final Set<Schema> unionSchemas = Collections.newSetFromMap(new IdentityHashMap<>());
+                // ArrayList ensures that ordering of subTypes is preserved.
+                final List<Schema> unionSchemas = new ArrayList<>();
                 // Initialize with this schema
                 if (_type.isConcrete()) {
                     unionSchemas.add(_typeSchema);
@@ -126,12 +125,26 @@ public class RecordVisitor
                 } catch (JsonMappingException jme) {
                     throw new RuntimeJsonMappingException("Failed to build schema", jme);
                 }
-                _avroSchema = Schema.createUnion(new ArrayList<>(unionSchemas));
+                _avroSchema = Schema.createUnion(deduplicateByReference(unionSchemas));
             } else {
                 _avroSchema = _typeSchema;
             }
         }
         _visitorWrapper.getSchemas().addSchema(type, _avroSchema);
+    }
+
+    private static List<Schema> deduplicateByReference(List<Schema> schemas) {
+        final List<Schema> result = new ArrayList<>();
+        // Set based on IdentityHashMap is used because we need to deduplicate by reference.
+        final Set<Schema> seenSchemas = Collections.newSetFromMap(new IdentityHashMap<>());
+
+        for(Schema s : schemas) {
+            if(!seenSchemas.contains(s)) {
+                seenSchemas.add(s);        // marks as seen by reference
+                result.add(s);             // preserve order
+            }
+        }
+        return result;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
```diff
@@ -98,9 +98,8 @@ public RecordVisitor(SerializerProvider p, JavaType type, VisitorFormatWrapperIm
...
                 // Therefore, unionSchemas must not be HashSet (or any other type
                 // using hashCode() for equality check).
-                // Set ensures that each subType schema is once in resulting union.
-                // IdentityHashMap is used because it is using reference-equality.
-                final Set<Schema> unionSchemas = Collections.newSetFromMap(new IdentityHashMap<>());
+                // ArrayList ensures that ordering of subTypes is preserved.
+                final List<Schema> unionSchemas = new ArrayList<>();
                 // Initialize with this schema
                 if (_type.isConcrete()) {
                     unionSchemas.add(_typeSchema);
```
cowtowncoder, line 105: Why not prevent duplicate adds here instead? Check if `unionSchemas` has exact instance (iterate over, compare with `==`), only add if not yet included?

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
```diff
@@ -126,14 +125,28 @@ public RecordVisitor(SerializerProvider p, JavaType type, VisitorFormatWrapperIm
...
         _visitorWrapper.getSchemas().addSchema(type, _avroSchema);
     }
 
+    private static List<Schema> deduplicateByReference(List<Schema> schemas) {
+        final List<Schema> result = new ArrayList<>();
+        // Set based on IdentityHashMap is used because we need to deduplicate by reference.
+        final Set<Schema> seenSchemas = Collections.newSetFromMap(new IdentityHashMap<>());
+
+        for(Schema s : schemas) {
+            if(!seenSchemas.contains(s)) {
```
cowtowncoder: If we were to use `Set`, instead of `set.contains()` should rather try to add, check response value (`true` means entry was added, i.e. was not yet included).

cowtowncoder:
Good start! PR much appreciated.
I think we can avoid separate de-duping by checking for existence on-the-fly, but aside from that I think this makes sense.

One process thing: we need CLA (from either https://github.com/FasterXML/jackson/blob/main/contributor-agreement.pdf (individual) or https://github.com/FasterXML/jackson/blob/main/contributor-agreement.pdf (Corporate CLA)). This is only needed once per contributor, one is good for all future contributions.

The usual way is to print, fill & sign, email to `cla` at fasterxml dot com. Once this is done I can proceed with merging (pending final review etc).

Looking forward to merging this PR (and backporting in 2.19, I think).

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
beseder1: Makes sense! Done.

avro/src/main/java/com/fasterxml/jackson/dataformat/avro/schema/RecordVisitor.java
beseder1: Thanks for the suggestion! I thought since there are actually 3 places in the code where schemas are being added to the `unionSchemas` list (in particular one with `addAll`) it would be a bit less cumbersome from a code perspective to simply de-duplicate at the end. If I am not getting something terribly wrong it seems to me that de-duplication at the end could be even better in terms of performance because when checking on the fly with `==` we could get 0 + 1 + 2 + ... + n many iterations, i.e. yielding O(n^2) complexity. De-duplication at the end using IdentityHashMap which provides constant time access for get/put on the other hand should rather yield O(n) complexity. What do you think?

cowtowncoder: Ah. Missed the code duplication aspect. 

I think we can go with what exists; it's not a big deal as schema construction should not be on critical path anyway for most usage.
Plus perf/big-O complexity is indeed unclear depending on schema size.

cowtowncoder (approved):
LGTM

cowtowncoder:
Ok ready to merge once we get CLA!

Thank you again @beseder1 .

cowtowncoder:
CLA received, will now merge

cowtowncoder (approved):

PR #614, opened 2025-09-22
Fixes #601: Jackson subtype schema unions are non-deterministic (#613)

### LINKED ISSUE #601: Jackson subtype Avro schema unions are non-deterministic and therefore incompatible with each other

beseder1:
Trying to cherry pick into 2.19


beseder1:
@cowtowncoder At least locally I am not getting any failing tests as opposed to your comment under https://github.com/FasterXML/jackson-dataformats-binary/issues/601.

beseder1:
Closing as # in branch name seems to cause issues with workflows not be executed.

PR #615, opened 2025-09-22
Fixes #601: Jackson subtype schema unions are non-deterministic (#613)

### LINKED ISSUE #601: Jackson subtype Avro schema unions are non-deterministic and therefore incompatible with each other

beseder1:
Trying to cherry pick into 2.19


beseder1:
I messed up the branch name in https://github.com/FasterXML/jackson-dataformats-binary/pull/614 hence closed and reopened a new PR. Let's see if the everything passes.

beseder1:
Okay seems that the pipeline succeeded and cherry picking into 2.19 works.

cowtowncoder (approved):

cowtowncoder:
Thank you @beseder1 ! Merged for 2.19(.3)

PR #568, opened 2025-03-09
Improve ASCII performance

sugmanue:
### Summary

Improve the ASCII case by creating a tight loop around it. All the changes follows a similar pattern. First attempt to do a tight loop around ASCII and fallback whenever a non-ascii char is found.

These changes shows improvements of up to 7x for the ASCII case, but also for the multi-byte code path.

The `_2` cases are for the same sizes but without chunking. The CBOR was created using [json2cbor](https://github.com/wormbks/cbor2json) to avoid chunking. All these benchmark test can be found [here](https://github.com/sugmanue/cbor-vs-json-perf-test)

### Benchmark Results

All the benchmarks can be found [here](https://github.com/sugmanue/cbor-vs-json-perf-test). 

```
Benchmark                (flavor)      (size)  Mode  Cnt       Score       Error  Units
MyBenchmark.cbor  ASCII_PRINTABLE       SMALL  avgt    5     283.767 ±     6.141  ns/op (before)
MyBenchmark.cbor  ASCII_PRINTABLE       SMALL  avgt    5     276.760 ±     4.900  ns/op (after)
MyBenchmark.cbor  ASCII_PRINTABLE      MEDIUM  avgt    5     751.102 ±     9.724  ns/op (before)
MyBenchmark.cbor  ASCII_PRINTABLE      MEDIUM  avgt    5     412.084 ±     0.906  ns/op (after)
MyBenchmark.cbor  ASCII_PRINTABLE       LARGE  avgt    5    1162.698 ±    34.077  ns/op (before)
MyBenchmark.cbor  ASCII_PRINTABLE       LARGE  avgt    5     537.463 ±     1.371  ns/op (after)
MyBenchmark.cbor  ASCII_PRINTABLE     X_LARGE  avgt    5   97592.433 ±   652.295  ns/op (before)
MyBenchmark.cbor  ASCII_PRINTABLE     X_LARGE  avgt    5   12433.531 ±    55.798  ns/op (after)
MyBenchmark.cbor  ASCII_PRINTABLE    XX_LARGE  avgt    5  192964.487 ±   764.024  ns/op (before)
MyBenchmark.cbor  ASCII_PRINTABLE    XX_LARGE  avgt    5   23451.347 ±    57.090  ns/op (after)
MyBenchmark.cbor  ASCII_PRINTABLE  XX_LARGE_2  avgt    5  192113.905 ±  1270.924  ns/op (before)
[...]

Issue #519, opened 2024-11-15
JSON vs CBOR performance for ASCII text


sugmanue:
Hi there,

We're testing the performance of CBOR vs plain JSON and looks like, at least for ASCII text, JSON is quite faster, this speaks volumes about the Jackson performance of JSON but looks like CBOR still has room for improvement. 

The performance tests can be found on [this repository](https://github.com/sugmanue/cbor-vs-json-perf-test). For a simple class with five `String` fields and ASCII (non-escaped) strings JSON is almost twice as fast as CBOR for larger strings (between 193 and 231 chars)

```
Benchmark                (flavor)  (size)  Mode  Cnt    Score    Error  Units
MyBenchmark.cbor  ASCII_PRINTABLE   SMALL  avgt    5  266.316 ±  9.938  ns/op
MyBenchmark.json  ASCII_PRINTABLE   SMALL  avgt    5  243.984 ± 13.422  ns/op

Benchmark                (flavor)  (size)  Mode  Cnt    Score    Error  Units
MyBenchmark.cbor  ASCII_PRINTABLE  MEDIUM  avgt    5  725.376 ±  8.700  ns/op
MyBenchmark.json  ASCII_PRINTABLE  MEDIUM  avgt    5  464.803 ± 20.404  ns/op

Benchmark                (flavor)  (size)  Mode  Cnt     Score    Error  Units
MyBenchmark.cbor  ASCII_PRINTABLE   LARGE  avgt    5  1126.297 ±  7.843  ns/op
MyBenchmark.json  ASCII_PRINTABLE   LARGE  avgt    5   664.466 ± 23.541  ns/op

```

As expected, this is not the case for multi-byte chars, for instance, chars from the CJK block, emojis or full ASCII (some of which requires escaping in plain JSON). See below

```
Benchmark         (flavor)  (size)  Mode  Cnt     Score    Error  Units
MyBenchmark.cbor       CJK   LARGE  avgt    5  2076.014 ± 67.569  ns/op
MyBenchmark.json       CJK   LARGE  avgt    5  2939.622 ± 16.501  ns/op

Benchmark         (flavor)  (size)  Mode  Cnt     Score     Error  Units
MyBenchmark.cbor     EMOJI   LARGE  avgt    5  2400.312 ±  11.203  ns/op
MyBenchmark.json     EMOJI   LARGE  avgt    5  8467.852 ± 243.559  ns/op

[...]

PR #617, opened 2025-09-29
Fix long ASCII text read

### LINKED ISSUE #616: CBOR text gets truncated on decoding
There are two bugs in the `_finishLongTextAscii` method introduced in #519 that produces the text to be truncated.

1. The `outPtr` is always set to zero (see [here](https://github.com/FasterXML/jackson-dataformats-binary/blob/b20075ff0c029d659cb24adc6c65d2be748a8753/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L2636)) before the read loop. If the out buffer still has room its contents will be overwritten instead of keep adding to it (overwrite or missing chunks.)
2. If the method exits by fully reading the expected text length, outside the [outer loop](https://github.com/FasterXML/jackson-dataformats-binary/blob/b20075ff0c029d659cb24adc6c65d2be748a8753/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L2657) the length of the last segment is not set, which makes the calling code to drop it when the string is finished (text truncated.)

sugmanue:
There are two bugs in the `_finishLongTextAscii` method introduced in #519 (via #568) that produces the text to be truncated.

1. The `outPtr` is always set to zero (see [here](https://github.com/FasterXML/jackson-dataformats-binary/blob/b20075ff0c029d659cb24adc6c65d2be748a8753/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L2636)) before the read loop. If the out buffer still has room its contents will be overwritten instead of keep adding to it (overwrite or missing chunks.)

2. If the method exits by fully reading the expected text length, outside the [outer loop](https://github.com/FasterXML/jackson-dataformats-binary/blob/b20075ff0c029d659cb24adc6c65d2be748a8753/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L2657) the length of the last segment is not set, which makes the calling code to drop it when the string is finished (text truncated.)

### Notes

1. This code path is only triggered by long texts, where long means that we cannot fully read its length in the input buffer. 
2. This change includes two tests, one for non-chunked text (the case being fixed here), and, another for chunked text to validate that this issue is not in that path as well.

Fixes #616.


cowtowncoder:
Whoa! Thank you very much for reporting #616 and providing this fix. I'll need to read it with thought.
I think we have CLA for you (as per #568) so that's good.

But due to nature of the bug, I think we'd want fix all the way to `2.18` branch (that's the intended LTS release. I could try cherry-picking, or, if it's easy enough for you, re-creating PR with target as `2.18` would be great.

One possibly gnarly change there is JUnit 4 -> 5 conversion (see #550), done for 2.19.

So alternatively could consider merging full PR in `2.19`, and only backporting fix, not tests (not ideal but... would do).

sugmanue:
> Whoa! Thank you very much for reporting #616 and providing this fix. I'll need to read it with thought. I think we have CLA for you (as per #568) so that's good.

I introduced it in the first place, somehow I didn't fully test it. I added Jacoco locally and verified that all the code introduced in the previous PR is covered. Apologies for my sloppiness.

Unrelated to this change, with Jacoco, I see some code paths related to reading numbers that are not covered. I wonder if there's any reason not add Jacoco? If there's none, I can send a PR for that.

sugmanue:
> But due to nature of the bug, I think we'd want fix all the way to `2.18` branch (that's the intended LTS release. I could try cherry-picking, or, if it's easy enough for you, re-creating PR with target as `2.18` would be great.

As far as I understand this change is only present on 2.19 onwards. I double checked the code present on 2.18 and this code path is not there. So, it's not affected by this particular issue.

cowtowncoder:
@sugmanue I should have checked before I wrote above: yes, this was changed in 2.19.0 so fix need not (and cannot) go in 2.18 anyway. But I think it'd be good to merge it in `2.19` just in case -- in case we'll release 2.19.3.

cowtowncoder:
@sugmanue np, these things happen. I did not review code well enough either. Glad it got caught now at least.

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
cowtowncoder, line 2640: (just noting for posterity, not suggesting change within this fix)

Check for `i >= 0` seems sub-optimally placed, before actually access and output of byte itself, leading to need to "undo" copy -- instead of changing control flow where problem encountered.

cowtowncoder (approved):
LGTM, will merge, backport

cowtowncoder:
Merged, backported in 2.19(.3), 2.20(.1).

Does NOT affect 2.18 (or earlier 2.x, I think).

cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
sugmanue: The idea was to remove as many branches from the loop as possible. The cost is the need to undo, but at least that branch is outside the hot code-path. I didn't do performance testing to validate the idea, but I will do some and post back the results.

cowtowncoder: yeah measuring is good -- actual performance is not always obvious. So in this case there's just one extra comparison (`i` checked before first copy) but that's only once per segment/run, probably insignificant over non-trivial data.

sugmanue: Benchmarks with the code as is (using the [this benchmarks](https://github.com/sugmanue/cbor-vs-json-perf-test)).

```
Benchmark                (flavor)    (size)  Mode  Cnt      Score     Error  Units
MyBenchmark.cbor  ASCII_PRINTABLE  XX_LARGE  avgt    5  24318.121 ± 329.952  ns/op
```

And with [this patch](https://gist.github.com/sugmanue/fa572b14c93e4180a69457de4f308523) applied.

```
Benchmark                (flavor)    (size)  Mode  Cnt      Score      Error  Units
MyBenchmark.cbor  ASCII_PRINTABLE  XX_LARGE  avgt    5  26899.778 ± 1239.516  ns/op
```

Looks like the current version is slightly faster, but not by much. This input has 4 fields of about 16Kb.

cowtowncoder: Ok. I'll take that. :)

Thank you for humoring me.

davidh44:
Hi, is there any timeline for releasing the new versions to Maven?

cowtowncoder:
@davidh44 Not really, more based on two main factors:

1. Number (and severity) of fixes accumulated for (open) branch
2. My time availability (limited by OSS not being my day job) as I need to do releases, takes ~2 hours when all goes well

Which version would you like to see, 2.19.3 or 2.20.1 ?

I will most likely need to first release 2.18.5; but then looking at accumulation, 2.20.1 probably more likely to be released just based on need.

davidh44:
I think either 2.19.3 or 2.20.1 would work, we want to pull in performance optimizations into the AWS SDK for Java v2 that were introduced in 2.19.0

We had to revert the bump due to this bug that this PR fixes
https://github.com/aws/aws-sdk-java-v2/commit/55adcf3e1b8e34582ec68d98256ddc19b3cc4c48

cowtowncoder:
@davidh44 Ok I'll try to get release of one if not both out this week: there are enough fixes and this is pretty critical one.

davidh44:
@cowtowncoder Awesome sounds good, thank you!

cowtowncoder:
Managed to release 2.18.5 today (see https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.18.5). Maybe 2.19.3 tomorrow. :)

PR #622, opened 2025-10-23
Update `ion-java` dep to `1.11.1` (from 1.11.10)

Shaurya0108:
## **Description**

Ion v1.11.11 ([release notes](https://github.com/amazon-ion/ion-java/releases/tag/v1.11.11)) changed how values that fit in long are now classified ([change in Ion Java](https://github.com/amazon-ion/ion-java/pull/1055)), instead of using `BIG INTEGER` they use `LONG`. Using a `BIG INTEGER` was causing test `testLongAsBigIntegerSize` to fail when the Ion version was changed to v1.11.11 from v1.11.10:

```
[ERROR] tools.jackson.dataformat.ion.IonNumberOverflowTest.testLongAsBigIntegerSize -- Time elapsed: 0.035 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <BIG_INTEGER> but was: <LONG>
```

The current version of Ion Java utilizes Jackson-dataformat-ion 2.18.

### DIFF
--- a/ion/pom.xml
+++ b/ion/pom.xml
@@ -38,7 +38,7 @@ tree model)
     <dependency>
       <groupId>com.amazon.ion</groupId>
       <artifactId>ion-java</artifactId>
-      <version>1.11.10</version>
+      <version>1.11.11</version>
     </dependency>
 
     <!-- will come from parent pom for later versions -->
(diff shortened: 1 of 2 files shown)


Shaurya0108:
@cowtowncoder would a release to 2.18 be possible? We are currently using 2.18.4 at Ion, and will not be able to move to versions 2.19 and greater for a bit.

cowtowncoder:
@Shaurya0108 Hmmh. Normally I would say no but... 2.18 is the first LTS version of Jackson and it seems like a small patch upgrade. So perhaps.

In the meantime I think you should be able to just override `ion-java` version being deployed, given only test changes were needed.

cowtowncoder:
I am thinking of possibly releasing 2.18.5 today, but we will see how timing goes (if I have time to do it).

Shaurya0108:
Thank you for taking the time to do so!

PR #625, opened 2025-10-26
Fix #624: use more efficient String constructor

### LINKED ISSUE #624: Use more efficient (in JDK 17) `String` construction wrt "Compact Strings" (for CBOR)
A major reason Jackson codecs are faster on JDK 8 than 17 or above is the switch by JDK to use `byte[]` for `String`s.
For Jackson this is very bad thing because it already decodes Strings into `char[]`.
With Jackson 3.0 baseline now is JDK 17 so maybe there are ways to figure out how to undo some damage, esp. for ASCII decoding loops.

cowtowncoder:
Implements #624.


cowtowncoder:
Oddly does not seem to speed up `jackson-benchmarks` CBOR test (in fact, ~5% slow down).

Maybe `new String(byte[], int, int, Charset)` not more optimal or... need to do some profiling I suppose.

cowtowncoder:
Oh. I suspect JDK uses fast method:

    byte[] bytes = StringLatin1.encode(chars, 0, chars.length);

to handle conversion.

cbor/src/main/java/tools/jackson/dataformat/cbor/CBORParser.java
```diff
@@ -2464,27 +2464,33 @@ private final String _finishShortText(int len) throws JacksonException
...
-        while (inPtr < end && i >= 0) {
-            i = inputBuf[inPtr++];
-            outBuf[outPtr++] = (char) i;
+        while (inPtr < end && inputBuf[inPtr] >= 0) {
+            ++inPtr;
         }
-        if (inPtr == end && i >= 0) {
-            String str = _textBuffer.setCurrentAndReturn(outPtr);
+        if (inPtr == end) { // all ASCII
+            // 25-Oct-2025: [dataformats-binary#624] Minor optimization: can directly
```
pjfanning, line 2479: @cowtowncoder the date is wrong

cowtowncoder: ah. Yes, from when I originally created PR. will fix.

Issue #629, opened 2025-10-30
Unnecessary `IOException` in `IonObjectMapper` method signatures


mcliedtke:
With the move to `RuntimeException`'s as a baseline in Jackson-3, there are some unnecessary `throws IOException` declarations in the method signatures of `readValue` and `writeValue` methods int he `IonObjectMapper` class. Could these `throws` declarations be removed?

There is one method - `writeValueAsIonValue` which technically does have an uncaught `IOException`, but given that it both opens and closes the writer to just in-memory objects, I don't see how an actual IOException would be anything other than an illegal state. I am not sure if there is a conventional way to handle this, but it would be nice to see this caught and rethrown as a `RuntimeException` base as well (is there a good corresponding JacksonException that fits? `JacksonIOException`?)

cowtowncoder:
+1 if I understand this correctly: `IOException` should no longer be exposed by anything in 3.x.
I think `src/main/java/tools/jackson/core/exc/JacksonIOException.java` is indeed the thing that `jackson-core` (and format modules) use.

Due to signature change, not 100% sure if this can be done for 3.0(.2) or does it need to go in 3.1(.0).
But +1 for doing this.

mcliedtke:
Let me know if this is roughly what is expected. https://github.com/FasterXML/jackson-dataformats-binary/pull/631

No strong opinions on 3.0(.2) vs. 3.1

cowtowncoder:
> No strong opinions on 3.0(.2) vs. 3.1

Right, that is more for me to consider, wrt compatibility. I think 3.1 it should be, so `3.x` is the target branch to use.

mcliedtke:
Thank you for the quick review and merge!

cowtowncoder:
Thank you for pointing out that missing part of 3.0 code conversion & fix!

PR #631, opened 2025-10-30
Remove IOException from IonObjectMapper read and write method signatures

mcliedtke:
Remove IOException from IonObjectMapper read and write method signatures. Most are unnecessary and one is now wrapped as JacksonIOException

Ref: https://github.com/FasterXML/jackson-dataformats-binary/issues/629


cowtowncoder:
Looks good, hoping to get merged later today.

PR #641, opened 2025-12-19
Fix a out of bounds exception when writing to the outBuf

### LINKED ISSUE #640: CBOR throws ArrayIndexOutOfBoundsException with long ASCII text
I found another issue with the long ASCII text code path. When the size equals the length of the `_outBuf` [here](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.20/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L2632) then the `_outBuf` will be try to be written to using an out of bound index, since the upper limit for the tight loop [down here](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.20/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java#L2640) is set to be `int outEnd = size + outPtr;`

The exception thrown is

```
java.lang.ArrayIndexOutOfBoundsException: Index 8538 out of bounds for length 8538

	at com.fasterxml.jackson.dataformat.cbor.CBORParser._finishLongTextAscii(CBORParser.java:2643)
```

The issue was introduced in #617. I will submit a small change to make sure that we don't attempt to write past the end of the `_outBuf`.

sugmanue:
Fix for a `ArrayIndexOutOfBoundsException` exception when parsing long non-chunked ASCII text. The issue is described in #640.

### Testing

I changed a bit the existing tests to account to allow the size to match to trigger the bug.

Fixes #640


cowtowncoder:
Not needed, merged 2.19 PR forward, closing.

PR #642, opened 2025-12-19
Fix a out of bounds exception when writing to the outBuf

### LINKED ISSUE #640: CBOR throws ArrayIndexOutOfBoundsException with long ASCII text

sugmanue:
Fix for a `ArrayIndexOutOfBoundsException` exception when parsing long non-chunked ASCII text. The issue is described in #640.

### Testing

I changed a bit the existing tests to account to allow the size to match to trigger the bug.

Fixes #640


cowtowncoder:
Since 2.18 is the LTS, will try to backport (cherry-pick failed). 2.19 no longer open, no new releases. 2.20 still open until 2.21 released.

**EDIT**: never mind -- forgot regression only in 2.19, 2.18 not affected. I should read descriptions with more care.

alextwoods:
@cowtowncoder  - is there an ECD/schedule for the next release of 2.19.5 that would include this fix?

cowtowncoder:
No more 2.19 releases planned (see https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.19).

Fix will be included in 2.20.2 (https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.20.2) and 2.21.0 (https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.21) -- latter more likely to be out first -- hopefully within week now.

cowtowncoder:
Fix was included in `2.20.2`, `2.21.0` and `3.0.4` versions that have now been released.

PR #687, opened 2026-04-07
Fix `_finishLongTextAscii` returning negative remaining length

### LINKED ISSUE #686: `CBORParser._finishLongTextAscii` returns negative length when `TextBuffer` segment > I/O buffer, leaving non-ASCII byte unconsumed
When upgrading from 2.18.6 to 2.19+, we noticed the following errors:
* `com.fasterxml.jackson.core.JsonParseException: Unsupported major type (5) for CBOR Objects, not (yet?) supported, only Strings`
* `com.fasterxml.jackson.core.JsonParseException: Unsupported major type (7) for CBOR Objects, not (yet?) supported, only Strings`
* `com.fasterxml.jackson.core.JsonParseException: Invalid 5-bit length indicator for JsonToken.VALUE_STRING: 0x1C; only 0x00-0x17, 0x1F allowed`

I asked Claude to figure out what's going on. It pointed out this line:

https://github.com/FasterXML/jackson-dataformats-binary/blob/560d5a3cdf238596f66e9830d356606520413e79/cbor/src/main/java/tools/jackson/dataformat/cbor/CBORParser.java#L2618

Claude summary:

`CBORParser._finishLongTextAscii` computes its return value as `len - outPtr`. When the TextBuffer's current segment is larger than the 8000-byte I/O buffer, `outPtr` can exceed `len` — producing a negative return value. The caller `_finishLongText` uses `while (--len >= 0)`, which never executes on a negative len, leaving the non-ASCII byte unconsumed in the input stream. The next CBOR token read misinterprets that byte, throwing e.g. `"Unsupported major type (5)" or "Non-string provided as map key"`. 

---

Affected versions: 2.19.x, 2.20.x, 2.21.x                                                                                              
Working version: 2.18.x

It seems https://github.com/FasterXML/jackson-dataformats-binary/commit/107c6fb3 is the cause.

hisener:
`return len - outPtr` was wrong when `outPtr` accumulated across multiple outer-loop iterations: `len` had already been decremented for previous iterations, so subtracting the full `outPtr` produced a negative value, causing `_finishLongText` to skip decoding the non-ASCII byte.

Fix: use `return len - _inputPtr`, which holds only the bytes consumed in the current iteration.

Fixes https://github.com/FasterXML/jackson-dataformats-binary/issues/686.

### DIFF
--- a/cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/parse/CBORLongAsciiRead686Test.java
+++ b/cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/parse/CBORLongAsciiRead686Test.java
@@ -13,7 +13,7 @@ import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 // For [dataformats-binary#686]
-public class CBORFinishLongTextAsciiBugTest
+public class CBORLongAsciiRead686Test
 {
     // TextBuffer segment sizes grow 1.5× per flip, starting at 200 chars (MIN=500):
     //     S0=200, S1=500, S2=750, S3=1125, S4=1687, S5=2530, S6=3795, S7=5692, S8=8538
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: ['8', '11', '17', '21']
+        java_version: ['8', '17', '21']
         include:
           - java_version: '8'
             release_build: 'R'
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -2620,7 +2620,8 @@ public class CBORParser extends ParserMinimalBase
     }
 
     /**
-     * Consumes as many ascii chars as possible in a tight loop. Returns the amount of bytes remaining.
+     * Consumes as many ascii chars as possible in a tight loop.
+     * Returns the amount of bytes remaining.
      */
     private final int _finishLongTextAscii(int len) throws IOException
     {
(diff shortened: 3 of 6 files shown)


cowtowncoder:
Ok, I think fix needs to against branch `2.21` since 2.18 and 2.21 are 2.x LTS versions and (2.18.x works).
I can merge forward from 2.x to 3.x but not vice versa.

cbor/src/test/java/com/fasterxml/jackson/dataformat/cbor/parse/CBORLongAsciiRead686Test.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
+import org.junit.jupiter.api.Test;
+
+import tools.jackson.databind.ObjectMapper;
+import tools.jackson.dataformat.cbor.*;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
```
cowtowncoder, line 14: Please add a comment referencing issue fixed:

```
// For [https://github.com/FasterXML/jackson-dataformats-binary/issues/686]
```

cowtowncoder:
Aside from re-creating PR against `2.21`, one process thing is that unless you have already sent it,
we'd need CLA (https://github.com/FasterXML/jackson/blob/main/CLA-jackson-2026.pdf).
This only needs to be sent once, and the usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com.

Once we have it, I can proceed with merging & CLA is good for all future contributions.

Thank you again -- looking forward to merging the fix!

hisener:
Addressed the comment and rebased onto 2.21. Also sent the CLA email.

cowtowncoder:
Thank you, @hisener ! Fix merged in, will be in 2.21.3 and 3.1.2.

hisener:
@cowtowncoder Is there a planned date for the 2.21.3 release?

cowtowncoder:
@hisener Depends on various things, including my availability (takes 2 hours of time to publish). Within next couple of weeks, maybe?

hisener:
@cowtowncoder sounds good; thanks!

PR #691, opened 2026-05-12
Add parameterized tests covering all ASCII-optimization exit paths in CBORParser

sugmanue:
### Summary

These test were missing from the original pull request #568, and caused that at least three bugs were found that have been missed in the original implementation due to the lack of test covering all corner cases (#687, #617, and, #640). Adding now to ensure that no regressions are introduced.

Each test generates a string with a controlled ASCII prefix followed by unicode, encodes it to CBOR, then decodes via both byte[] and throttled InputStream to exercise both _finishShortText and _finishLongTextAscii/_finishChunkedTextAscii paths.

Since the new tests covers all known cases the previously added test `ParseLongAsciiTextTest.java` was removed as it's now redundant alongside with the resources it used.

Also two small changes in CBORParser

* A performance improvement to call `_loadMoreForChunkIfNeeded()` only called when we've actually exhausted the current window
* A change to reset `_textBuffer`, this code path will already throw if we cannot read enough but it's still worth leaving the `_textBuffer` in a correct state.


cowtowncoder:
Thank you, @sugmanue. Will get this reviewed, merged.

cowtowncoder (approved):
LGTM (made some minor clean up)

Issue #693, opened 2026-05-26
Incomplete number length validation in Avro decoder (for `BigDecimal`)


cowtowncoder:
(original report by @tonghuaroot)

Checks for `StreamReadConstraints.maxNumberLength` are incomplete for Avro decoder wrt types `BigDecimal` and `BigInteger`.
All 2.x and 3.x versions appear affected: if so, should be fixed in 2.18 (oldest LTS), merged forward.

**EDIT**: actually no, only 2.19 and above have `AvroParserImpl.decodeBytesDecimal()` and related methods. So fix should go in `2.21`.

PR #695, opened 2026-05-26
[avro] Apply StreamReadConstraints to BigDecimal decode paths (#693, 3.x)

tonghuaroot:
## Summary

3.x mirror of #694 for #693.

`AvroParserImpl.decodeBytesDecimal` and `decodeFixedDecimal` did not invoke
`StreamReadConstraints.validateFPLength` against the raw 2's-complement
payload length before constructing the `BigDecimal` value. Sibling binary
parsers (`CBORParser`, `SmileParser`) cap via byte-length at the
equivalent native big-number token. This change aligns `AvroParserImpl`
with that shape on 3.x.

## Changes

- `avro/src/main/java/tools/jackson/dataformat/avro/deser/AvroParserImpl.java`:
  apply `streamReadConstraints().validateFPLength(_binaryValue.length)` on
  both the `bytes`-backed and `fixed`-backed decimal decode paths.
- `avro/src/test/java/tools/jackson/dataformat/avro/dos/AvroDecimalLengthLimitTest.java`:
  verifies that values larger than the configured `maxNumberLength` are
  rejected for both decimal payload forms, and that small / default-limit
  decimals continue to round-trip.

## Test plan

`mvn -pl avro test`

- 152 tests, 0 failures (148 existing + 4 new).
- Without the patch, the two `*LargePayloadTripsLimit` tests fail
  (no exception thrown); the two non-limit tests pass.
- With the patch, all four new tests pass.

### DIFF
--- a/avro/src/main/java/tools/jackson/dataformat/avro/deser/AvroParserImpl.java
+++ b/avro/src/main/java/tools/jackson/dataformat/avro/deser/AvroParserImpl.java
@@ -627,6 +627,8 @@ public abstract class AvroParserImpl
     // @since 2.19
     public JsonToken decodeBytesDecimal(int scale) throws IOException {
         decodeBytes();
+        // Enforce maxNumberLength against raw 2's-complement payload length,
+        // matching CBOR/Smile codecs handling which also uses raw byte length.
         streamReadConstraints().validateFPLength(_binaryValue.length);
         _numberBigDecimal = new BigDecimal(new BigInteger(_binaryValue), scale);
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Thank you @tonghuaroot !

cowtowncoder:
Ok, going forward, we'll need backported fixed for 3.1 whenever something is merged to 2.x LTS version(s).

cowtowncoder:
Merging from 2.x, similar to this PR; closing this one.

PR #710, opened 2026-07-10
Use `ArrayDeque` instead of (synchronized) `Stack` in CBORParser's string-ref tracking

benoitmeriaux:
`CBORParser`'s internal `StringRefListStack` (added for CBOR string-ref tag
support, per its javadoc `@since 2.15`) is backed by `java.util.Stack`.
`Stack` extends `Vector`, so every one of its methods is `synchronized`.

This patch replaces the backing `Stack<StringRefList>` with an unsynchronized
`ArrayDeque<StringRefList>`. Behavior is unchanged: `push`/`pop`/`peek` are
used purely as a LIFO stack, which `ArrayDeque` supports directly, and the
public `empty()` method is kept (now delegating to `isEmpty()`) so no caller
in `CBORParser` or downstream needs to change.

### Why

`StringRefListStack.push()`, `.pop()`, and `.empty()` are called on **every**
object/array boundary and on **every** string/property-name decode
(`_decodeChunkLength`, `_decodePropertyName`, `_finishTextToken`, etc.),
regardless of whether the document being parsed actually uses the CBOR
string-ref extension (tags 25/256) at all. Because the backing collection is
a `Stack`, each of those calls pays a `synchronized` monitor enter/exit, even
though a given `CBORParser` instance is never shared across threads and there
is no contention to guard against.

On string/property-name-heavy CBOR workloads this synchronization overhead
is measurable: in a production JFR/async-profiler CPU profile of a service
deserializing CBOR-encoded event data, `Vector.size()` (called internally by
`Stack.empty()`/`peek()`) and `Stack.empty()` itself showed up as some of the
hottest standard-library frames under `CBORParser`, ahead of most of the
actual decoding work.

`ArrayDeque` provides the exact same single-threaded LIFO semantics used here
(`push`/`pop`/`peek`/`isEmpty`) without any locking.

### Testing

- Existing CBOR test suite passes unchanged.
- Manually round-tripped deeply nested documents (50+ levels of
  object/array nesting) and documents with `CBORGenerator.Feature.STRINGREF`
  explicitly enabled and repeated field names/values, to exercise the
  push/pop/peek depth-tracking path this touches. Output is byte-for-byte
  identical to before the change, and decoded values are unchanged.
- Added `cbor/src/test/java/tools/jackson/dataformat/cbor/bench/StringRefListStackBench.java`, a standalone micro-benchmark reproducing both implementations under a representative push/peek/pop pattern, single- and multi-threaded. With default JIT settings it shows no difference (escape analysis elides the locks in the synthetic loop). Run with -XX:-DoEscapeAnalysis to reproduce the effect seen in profiling: Stack is ~3.2-3.6x slower than ArrayDeque there, consistently across 1/2/4/16 threads.

### Notes

- I didn't find an existing tracked issue for this — happy to file one and
  reference it in the changelog/comment if that's the preferred process
  before merging.
- No public API changes; this is purely an internal implementation detail of
  `CBORParser`.

### DIFF
--- a/cbor/src/main/java/tools/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/tools/jackson/dataformat/cbor/CBORParser.java
@@ -9,7 +9,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Stack;
+import java.util.ArrayDeque;
 
 import tools.jackson.core.*;
 import tools.jackson.core.base.ParserBase;
@@ -206,7 +206,7 @@ public class CBORParser extends ParserBase
 
         public void pop() {
             --_nestedDepth;
-            if (!_stringRefs.empty() && _stringRefs.peek().depth == _nestedDepth) {
+            if (!_stringRefs.isEmpty() && _stringRefs.peek().depth == _nestedDepth) {
                 _stringRefs.pop();
             }
         }
@@ -216,10 +216,10 @@ public class CBORParser extends ParserBase
         }
 
         public boolean empty() {
-            return _stringRefs.empty();
+            return _stringRefs.isEmpty();
         }
 
-        private Stack<StringRefList> _stringRefs = new Stack<>();
+        private ArrayDeque<StringRefList> _stringRefs = new ArrayDeque<>();
         private int _nestedDepth = 0;
     }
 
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Sounds good, thank you -- PR is enough no need for issue.

The only thing (beyond my needing to review) is that unless we have CLA (https://github.com/FasterXML/jackson/blob/main/CLA-jackson-2026.pdf) from you, we'd need it now; it's good for all future PRs.
The usual way is to print, fill & sign, email to `cla` at fasterxml dot com and once we get it we are good to go.

Looking forward to merging this, thanks again!

cowtowncoder:
Will probably backport in 3.1 -> 3.2 as well.

cowtowncoder:
Merged via #743 instead (for #742) -- thank you @benoitmeriaux for suggesting!

PR #721, opened 2026-07-17
Fix #720: (smile) SmileGenerator int overflow in maxLen for very long Strings

### LINKED ISSUE #720: `SmileGenerator` writes past output buffer (`ArrayIndexOutOfBoundsException`) for very long Strings due to int overflow in maxLen
### Affected version(s)

Reproduced with `jackson-dataformat-smile` **2.21.4**. The offending code is still present on the **2.20** branch and `master` (see line references below), so all current 2.x releases are affected.

### Summary

`SmileGenerator`, when serializing a `String` value (or field name) whose length exceeds `~715.8 million` characters, throws:

```
java.lang.ArrayIndexOutOfBoundsException: Index 8000 out of bounds for length 8000
    at com.fasterxml.jackson.dataformat.smile.SmileGenerator._shortUTF8Encode(SmileGenerator.java:1984)
    at com.fasterxml.jackson.dataformat.smile.SmileGenerator._writeNonSharedString(SmileGenerator.java:1065)
    at com.fasterxml.jackson.dataformat.smile.SmileGenerator.writeString(SmileGenerator.java:988)
    at com.fasterxml.jackson.databind.ser.std.StringSerializer.serializeContents(StringSerializer.java:40)
    ...
```

(`8000` is the default `BufferRecycler` write-encoding buffer size, so the exact number varies with buffer configuration.)

### Root cause

In `_writeNonSharedString`:

```java
private final void _writeNonSharedString(final String text, final int len) throws IOException {
    // Expansion can be 3x for Unicode; and then there's type byte and end marker, so:
    int maxLen = len + len + len + 2;          // <-- int arithmetic
    // Next: does it always fit within output buffer?
    if (maxLen > _outputBuffer.length) {       // guard: route big strings to the chunked encoder
        _writeByte(SmileConstants.TOKEN_MISC_LONG_TEXT_UNICODE);
        _mediumUTF8Encode(text, 0, len);       // safe, chunked path
        _writeByte(BYTE_MARKER_END_OF_STRING);
        return;
    }
    if ((_outputTail + maxLen) >= _outputEnd) {
        _flushBuffer();
    }
    ...
    int byteLen = _shortUTF8Encode(text, 0, len);   // assumes the whole string fits in the buffer
    ...
}
```

`maxLen = len + len + len + 2` is computed with 32-bit `int` arithmetic. When
`len >= 715_827_882` (i.e. `3*len + 2 > Integer.MAX_VALUE`), `maxLen` overflows to a
**negative** value. That defeats both guards:

- `maxLen > _outputBuffer.length` is `false` (a negative number is not `> 8000`), so the safe chunked `_mediumUTF8Encode` path is skipped.
- `(_outputTail + maxLen) >= _outputEnd` is also `false`, so no flush/room check happens.

Execution falls through to `_shortUTF8Encode(text, 0, len)`, whose tight ASCII loop
(`outBuf[ptr++] = (byte) c;`) writes the entire string into the fixed-size
`_outputBuffer`, overrunning it and throwing `ArrayIndexOutOfBoundsException` at the
buffer boundary.

For `len < 715_827_882`, `maxLen` is correct and the short path is only taken when
`3*len + 2 <= _outputBuffer.length` (≤ 2666 chars), which physically cannot overflow —
so the bug is strictly limited to the integer-overflow regime.

### Other affected sites (same pattern)

The same `len + len + len (+ 2)` int computation appears in:

- `_writeNonSharedString` — `SmileGenerator.java:1049` (2.20 branch)
- `writeFieldName(...)` long-name path — `SmileGenerator.java:814`
- `writeString(char[], int, int)` long-string path — `SmileGenerator.java:1110`
- raw UTF-8 write path — `SmileGenerator.java:1218`

All four should be reviewed for the same overflow.

> Note: this is distinct from the recently-fixed #640 / #641 / #642 / #643, which are on
> the **CBOR parser** (read) path (`CBORParser._finishLongTextAscii`), and from #259
> (CBOR parser handling of a ~`Integer.MAX_VALUE` alleged length). This report is the
> **Smile generator** (write) path.

### Suggested fix

Compute `maxLen` in `long` (or explicitly check `len` against a safe threshold) before
comparing to the buffer length, so oversized strings are routed to `_mediumUTF8Encode`:

```java
long maxLen = (long) len + len + len + 2;
if (maxLen > _outputBuffer.length) {
    // chunked path
    ...
    return;
}
```

Apply the same guard to the other three sites listed above.

### Minimal reproduction

Requires a large heap (the `String` alone is ~1.4 GB in memory as UTF-16). Run with e.g. `-Xmx4g`:

```java
import com.fasterxml.jackson.dataformat.smile.SmileFactory;
import java.io.OutputStream;
import java.nio.file.*;

public class SmileBigStringRepro {
    public static void main(String[] args) throws Exception {
        int len = 715_827_882;                 // 3*len + 2 overflows int -> negative maxLen
        char[] chars = new char[len];
        java.util.Arrays.fill(chars, 'a');     // all-ASCII to hit the fast loop
        String big = new String(chars);

        ObjectMapper mapper = new ObjectMapper(new SmileFactory());
        try (OutputStream os = Files.newOutputStream(Paths.get("out.smile"))) {
            mapper.writeValue(os, big);        // throws ArrayIndexOutOfBoundsException
        }
    }
}
```

### How we hit it in the wild

Serializing an OpenRewrite `org.openrewrite.text.PlainText` LST whose `text` field was a
single generated file larger than ~700 MB. Standard `ObjectMapper.writerFor(...).writeValue(outputStream, value)`
with a `SmileFactory`; no custom generator usage. The failure is fully explained by the
overflow above — no concurrency or buffer-reuse required.

bmuschko:
Fixes #720.

## Problem

`SmileGenerator` computes `maxLen = len + len + len (+ 2)` with 32-bit `int` arithmetic when deciding whether an encoded String fits in the output buffer. For a String (or field name) whose length is `>= 715_827_882`, `3*len + 2` overflows to a **negative** value, which defeats the "does it fit in the buffer?" guard. Execution falls through to the fast in-buffer encoder (`_shortUTF8Encode`), which then overruns the fixed-size `_outputBuffer` and throws:

```
java.lang.ArrayIndexOutOfBoundsException: Index 8000 out of bounds for length 8000
    at ...SmileGenerator._shortUTF8Encode(...)
    at ...SmileGenerator._writeNonSharedString(...)
    at ...SmileGenerator.writeString(...)
```

## Fix

Compute `maxLen` in `long` at all four sites where this guard appears, so oversized values are correctly routed to the chunked `_mediumUTF8Encode` path:

- `_writeNonShortFieldName` (long field-name path)
- `_writeNonSharedString` (`writeString(String)` — the reported stack)
- `writeString(char[], off, len)` (long char[] string path)
- `writeUTF8String(byte[], off, len)` (raw UTF-8 write path)

## Test

Adds `SmileGeneratorHugeStringTest`, a regression test covering the `String` and `char[]` value paths. A genuine reproduction needs a ~1.4 GB `char[]`, which exceeds the module's default `-Xmx1024m` test heap, so the test **self-skips via `assumeTrue`** in normal CI and runs only when a large heap is available (verified with `-Xmx6g`).

Verified that reverting the fix makes this test reproduce the exact `ArrayIndexOutOfBoundsException: Index 8000 out of bounds for length 8000` from the report, and the fix makes it pass.

Note: no `release-notes/VERSION` or `release-notes/CREDITS` entries are included in this PR.

### DIFF
--- a/smile/src/main/java/tools/jackson/dataformat/smile/SmileGenerator.java
+++ b/smile/src/main/java/tools/jackson/dataformat/smile/SmileGenerator.java
@@ -662,7 +662,7 @@ public class SmileGenerator
         _writeByte(TOKEN_KEY_LONG_STRING);
         // can we still make a temp copy?
         // but will encoded version fit in buffer?
-        int maxLen = len + len + len;
+        long maxLen = (long) len + len + len;
         if (maxLen <= _outputBuffer.length) { // yes indeed
             if ((_outputTail + maxLen) >= _outputEnd) {
                 _flushBuffer();
@@ -905,7 +905,7 @@ public class SmileGenerator
     private final void _writeNonSharedString(final String text, final int len) throws JacksonException
     {
         // Expansion can be 3x for Unicode; and then there's type byte and end marker, so:
-        int maxLen = len + len + len + 2;
+        long maxLen = (long) len + len + len + 2;
         // Next: does it always fit within output buffer?
         if (maxLen > _outputBuffer.length) { // nope
             // can't rewrite type buffer, so can't speculate it might be all-ASCII
@@ -965,7 +965,7 @@ public class SmileGenerator
             _outputBuffer[origOffset] = typeToken;
         } else { // "long" String, never shared
             // but might still fit within buffer?
-            int maxLen = len + len + len + 2;
+            long maxLen = (long) len + len + len + 2;
             if (maxLen <= _outputBuffer.length) { // yes indeed
                 if ((_outputTail + maxLen) >= _outputEnd) {
                     _flushBuffer();
@@ -1075,7 +1075,7 @@ public class SmileGenerator
             }
         } else { // "long" String
             // but might still fit within buffer?
-            int maxLen = len + len + len + 2;
+            long maxLen = (long) len + len + len + 2;
             if (maxLen <= _outputBuffer.length) { // yes indeed
                 if ((_outputTail + maxLen) >= _outputEnd) {
                     _flushBuffer();
--- a/smile/src/test/java/tools/jackson/dataformat/smile/gen/SmileGeneratorHugeStringTest.java
+++ b/smile/src/test/java/tools/jackson/dataformat/smile/gen/SmileGeneratorHugeStringTest.java
@@ -1,0 +1,62 @@
+package tools.jackson.dataformat.smile.gen;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Arrays;
+
+import org.junit.jupiter.api.Test;
+
+import tools.jackson.core.JsonGenerator;
+import tools.jackson.dataformat.smile.BaseTestForSmile;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+public class SmileGeneratorHugeStringTest extends BaseTestForSmile
+{
+    // 3*len + 2 overflows a signed int for this length
+    private final static int OVERFLOW_LEN = 715_827_882;
+
+    // Require a comfortable margin of heap before even attempting allocation
+    private final static long REQUIRED_HEAP = 3L * 1024 * 1024 * 1024; // 3 GB
+
+    // Discards output but counts bytes, so we do not also retain ~700 MB of encoded data
+    private static final class CountingOutputStream extends OutputStream {
+        long count;
+
+        @Override
+        public void write(int b) { count++; }
+
+        @Override
+        public void write(byte[] b, int off, int len) { count += len; }
+    }
+
+    @Test
+    public void testHugeAsciiStringDoesNotOverflowBuffer() throws IOException
+    {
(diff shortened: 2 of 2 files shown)


cowtowncoder:
Makes sense, thank you for submitting this!

One process thing: if we haven't gotten one (one is all we need!), we'd need CLA:

https://github.com/FasterXML/jackson/blob/main/CLA-jackson-2026.pdf

and the usual way is to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.

Once that is in I can do final review, merge & backport this in `2.21` / `3.1` as well (unless you want to re-create against `2.21` -- if you do I can merge forward easily; not a big deal either way).

Thank you once again @bmuschko !

bmuschko:
I am sorry but at the moment I cannot sign a CLA. I am hoping that you may commit the change through your account.

cowtowncoder:
@bmuschko No problem, I should be able to. Thank you for showing the way.

bmuschko:
Thank you!

bmuschko:
Actually fix is here: https://github.com/FasterXML/jackson-dataformats-binary/pull/722

PR #722, opened 2026-07-17
Fix #720: fail to write very long Smile strings

### LINKED ISSUE #720: `SmileGenerator` writes past output buffer (`ArrayIndexOutOfBoundsException`) for very long Strings due to int overflow in maxLen

cowtowncoder:


bmuschko:
Do you know when this fix will be released?

cowtowncoder:
@bmuschko Not exactly but hoping within next 2 weeks. Problem being we have 5 patches to make and each takes about 2 hours -- so it's a non-trivial commitment.

bmuschko:
I see that you created some releases last week. Where can I see the release notes for changes? Are those usually generated so I can identify when this change has been shipped?

cowtowncoder:
@bmuschko https://github.com/FasterXML/jackson/wiki/Jackson-Releases from which you get to branches (like https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.1) and patches (https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.1.6).

And yes, I completed the 5 version release set yesterday; managed to bring time per release to under 1.5 hours.

PR #738, opened 2026-07-25
Fix #735: CBOR stringref property-name paths pass length marker instead of actual length

### LINKED ISSUE #735: "stringref" property-name paths pass 5-bit length marker instead of actual length to `shouldReferenceString()`
### Problem

On the Object property name paths, `CBORParser` decides whether a name goes into the
CBOR "stringref" (http://cbor.schmorp.de/stringref) reference table by passing the
**5-bit length marker** to `shouldReferenceString()`, where the actual byte length of
the name is required:

2.18 (`CBORParser.java`, lines ~1573 in the `nextFieldName()` fast path and ~2949 in
`_decodePropertyName()`); same two sites on 2.19 - 2.x:

```java
        if (!chunked && !_stringRefs.empty() &&
                shouldReferenceString(_stringRefs.peek().stringRefs.size(), lenMarker)) {
            _stringRefs.peek().stringRefs.add(name);
            _sharedString = name;
        }
```

For markers 0 - 23 the marker *is* the length, so those are correct. For markers
24 - 27 ("1/2/4/8-byte length suffix follows") the value passed is literally 24, 25,
26 or 27 rather than the name's length.

All of the value paths (`_finishShortText()`, `_finishLongText()`, `_finishBytes()`)
pass the real byte length, so names and values disagree within the same parser.

### Why it matters

`CBORConstants.shouldReferenceString(int index, int stringBytes)` requires:

| Reference index | Minimum length |
| --- | --- |
| <= 23 | 3 bytes |
| <= 255 | 4 bytes |
| <= 65535 | 5 bytes |
| >= 65536 | 7 bytes |

Markers 24 - 27 are all >= 7, so they satisfy every threshold unconditionally. The
parser therefore registers a name that a conformant encoder would have skipped
whenever the name's real length is below the threshold for the current index but it
was encoded with a **non-minimal length prefix** — e.g. marker 24 with length 2
(`0x78 0x02 'a' 'b'`), which is legal-but-non-canonical CBOR.

That adds a phantom entry, so the parser's table is off by one from the encoder's from
that point on, and every following reference resolves to the wrong String (or fails
with "String reference out of range").

Canonically encoded names cannot trigger it: marker 24 implies length >= 24, marker 25
length >= 256, and so on, all above every threshold. So this only bites on
non-canonically encoded documents — hence low severity, but it is still a decoding
divergence from the spec driven purely by how a third-party encoder chose to write its
length prefixes.

### Note on 3.x

3.x has three name-path call sites, and they are inconsistent with each other:

* `tools/.../CBORParser.java` ~1515 (in `_decodePropertyName2()`) correctly uses
  `actualLen`
* ~1443 (`nextName(PropertyNameMatcher)`) and ~3125 use `lenMarker`

The one at ~1443 is a slightly different use — it decides whether to take the
off-line path at all rather than whether to register — but it should still be
reasoning about the real length.

### Versions

Stringref support was added in 2.15.0 (#347) and these call sites have had this shape
since, so 2.15 - 2.x and all of 3.x.

### Suggested fix

Pass the decoded length (`actualLen`, as 3.x's `_decodePropertyName2()` already does)
rather than `lenMarker`, at every name-path call site.

arimu1:
Fixes https://github.com/FasterXML/jackson-dataformats-binary/issues/735

### Problem

On the Object property-name decode paths, `CBORParser` decided whether a
name should be added to the CBOR "stringref" reference table by passing
the raw 5-bit length marker to `shouldReferenceString()` instead of the
name's actual decoded byte length:

- `nextFieldName()` fast path
- `_decodePropertyName()`

For markers 0-23 the marker *is* the length, so those paths were correct.
For markers 24-27 ("1/2/4/8-byte length suffix follows") the marker value
itself (24, 25, 26, or 27) was passed on unconditionally. Since
`CBORConstants.shouldReferenceString()`'s thresholds top out at 7 bytes,
markers 24-27 satisfy every threshold regardless of the name's real
length.

That means a non-canonically encoded short name (legal-but-non-minimal
CBOR, e.g. marker 24 with an explicit length of 2, `0x78 0x02 'a' 'b'`)
gets registered in the parser's stringref table when a conformant encoder
would have skipped it (its real length is below the threshold for the
current index). The parser's table then desyncs from the encoder's from
that point on, so later backreferences resolve to the wrong String or
fail with "String reference out of range".

All the value paths (`_finishShortText()`, `_finishLongText()`,
`_finishBytes()`) already pass the real decoded length, so this brought
the two name paths in line with them.

### Fix

Both sites now track the actual decoded length (the explicit length read
from the 1/2/4/8-byte suffix when present) in a separate `nameLen`
variable and pass that to `shouldReferenceString()` instead of the raw
`lenMarker`.

### Testing

Added `StringrefTest#testNonCanonicalLengthPropertyName`, which encodes a
property name with a non-canonical length prefix (`0x78 0x02 'a' 'b'`,
the issue's example) inside a stringref namespace, followed by a longer
canonical name and a stringref backreference, and asserts the
backreference resolves to the correctly-registered name. Reverting the
fix locally reproduces the bug: the backreference resolves to `"ab"`
instead of `"longname"`.

Ran the CBOR module's full test suite (JDK 17): 459 tests, 0 failures.

Targeted `2.21` per this repo's branch model (oldest maintained 2.x line;
merges forward through `2.22` -> `2.x` -> `3.1` -> ... automatically).

### DIFF
--- a/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
+++ b/cbor/src/main/java/com/fasterxml/jackson/dataformat/cbor/CBORParser.java
@@ -1614,6 +1614,7 @@ public class CBORParser extends ParserMinimalBase
             _sharedString = null;
             String name;
             boolean chunked = false;
+            int nameLen = lenMarker;
             if (lenMarker <= 23) {
                 // NOTE: [dataformats-binary#725] `maxNameLength` NOT enforced for
                 //   these shortest (at most 23 bytes) names; see
@@ -1646,10 +1647,11 @@ public class CBORParser extends ParserMinimalBase
                     //    decoding (or even reading) content
                     _streamReadConstraints.validateNameLength(actualLen);
                     name = _decodeLongerName(actualLen);
+                    nameLen = actualLen;
                 }
             }
             if (!chunked && !_stringRefs.empty() &&
-                    shouldReferenceString(_stringRefs.peek().stringRefs.size(), lenMarker)) {
+                    shouldReferenceString(_stringRefs.peek().stringRefs.size(), nameLen)) {
                 _stringRefs.peek().stringRefs.add(name);
                 _sharedString = name;
             }
@@ -3123,6 +3125,7 @@ CBORConstants.MAJOR_TYPE_BYTES, type);
         final int lenMarker = ch & 0x1F;
         boolean chunked = false;
         String name;
+        int nameLen = lenMarker;
         if (lenMarker <= 23) {
             // NOTE: [dataformats-binary#725] `maxNameLength` NOT enforced for
             //   these shortest (at most 23 bytes) names: enforcement is
@@ -3156,10 +3159,11 @@ CBORConstants.MAJOR_TYPE_BYTES, type);
                 //    decoding (or even reading) content
                 _streamReadConstraints.validateNameLength(actualLen);
                 name = _decodeLongerName(actualLen);
+                nameLen = actualLen;
             }
         }
         if (!chunked && !_stringRefs.empty() &&
-                shouldReferenceString(_stringRefs.peek().stringRefs.size(), lenMarker)) {
+                shouldReferenceString(_stringRefs.peek().stringRefs.size(), nameLen)) {
             _stringRefs.peek().stringRefs.add(name);
             _sharedString = name;
         }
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Sorry, did not see this and generated #739 instead.

PR #745, opened 2026-08-13
Fix avro bugs

pjfanning:
Claude AI found these - they look plausible to me but I must admit to not being an Avro expert.

Fix 1: _resolveBigDecimalIndex — dead code / wrong union branch

AvroWriteContext.java:441-448 — The second if (t == Type.DOUBLE) was dead code (unreachable after the first). The intent was to prefer STRING/BYTES (which preserve BigDecimal precision via logical types) and fall back to DOUBLE. Fixed to check STRING and BYTES first, with DOUBLE as the fallback match.

Fix 2: Stale type variable in both _createRecord overloads

AvroWriteContext.java:157-168, 179-195 — Both overloads captured type before union resolution but never updated it afterward. The MAP guard checked the stale type (still UNION) and would never trigger if the resolved schema was a MAP. Fixed by re-reading type = schema.getType() after union resolution.

Fix 3: Integer defaults stored as float — precision loss

AvroFieldDefaulters.java:31-38 — VALUE_NUMBER_INT cases used FloatDefaults (24-bit mantissa), silently losing precision for int values > 16,777,216 and all long values. IntDefaults and LongDefaults already existed in ScalarDefaults.java but were unused. Fixed to use IntDefaults for INT and LongDefaults for LONG/BIG_INTEGER.


avro/src/test/java/tools/jackson/dataformat/avro/BugFixTest.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Regression tests for bugs found during code analysis.
+ */
+public class BugFixTest extends AvroTestBase
```
cowtowncoder: We need better name here; may want to split across multiple test classes (ideally pre-existing).

pjfanning: I moved the tests

avro/src/test/java/tools/jackson/dataformat/avro/BigDecimal_serialization_and_deserializationTest.java
cowtowncoder: Class name should follow naming convention, no underscores.

cowtowncoder:
Looks good, but does require entry in `release-notes/VERSION-2.x`, something like:

```
#745: (avro) Misc schema bug fixes
```

(or ideally more specific -- but something)

avro/src/test/java/tools/jackson/dataformat/avro/BigDecimal_serialization_and_deserializationTest.java
pjfanning: this is a pre-existing class - do you want me to rename the file?

pjfanning: I renamed the class but there are other classes in this repo that have this non-standard naming

pjfanning:
> Looks good, but does require entry in `release-notes/VERSION-2.x`, something like:
> 
> ```
> #745: (avro) Misc schema bug fixes
> ```
> 
> (or ideally more specific -- but something)

Added an entry to VERSION file - this PR is targeted at 3.x so I added the entry under 3.3.0.

avro/src/test/java/tools/jackson/dataformat/avro/BigDecimal_serialization_and_deserializationTest.java
cowtowncoder: Ah sorry, did not realize. Thank you for renaming, going forward will try to keep an eye on additions.

cowtowncoder:
Hmmh. Claude found many issues here, will need to massage a bit...

PR #749, opened 2026-08-13
Improve avro `ObjectWriteContext` field lookup perf

pjfanning:
The fix builds a Map<String, Schema.Field> once in the constructor (O(n) at construction time) and uses Map.get() for O(1) lookups in both writeName() and _findField(), instead of calling _schema.getField(name) which does a linear scan on every field write.


cowtowncoder:
I am guessing there is some limit after which this is a win -- for small Maps overhead higher than linear lookup cost.
So would it make sense to gate it to some property size? Only do maps for 4+ entries f.ex?

Ideally small micro-benchmark to test there's speed up. Actually, we'd need something to show it's worth it for at least some cases; building map is not free so it may not speed things up for common cases.

pjfanning:
I'm putting this on the backburner, probably only really helps in edge cases

PR #750, opened 2026-08-13
SmileGenerator: avoid String allocation in `SmileGenerator.writeString(char[],int,int)`

pjfanning:
suggested by Claude AI

- Before: Always allocates new String(text, offset, len) to delegate to writeString(String), even for non-shared strings
- After: Uses _findSeenStringValue(char[], int, int) to look up by char array (no allocation). Only allocates a String when storing into the shared-value table (uncommon — only for strings that will be referenced again)


cowtowncoder:
Sounds reasonable; hoping to review soon.

cowtowncoder:
Had Claude do some verification, actual perf test; changes as result.

Hoping to finalize tomorrow, merge.

PR #752, opened 2026-08-13
Use varhandle for array access in `CBORGenerator`

pjfanning:
The VarHandle path is a single store instruction on most JVMs, versus 8 separate shift-cast-store operations. The same utility could be shared with the Smile module if desired.

Fallback to existing code if var handles are not supported, eg Android.


cbor/src/main/java/tools/jackson/dataformat/cbor/CBORVarHandleUtil.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+import java.nio.ByteOrder;
+
+/**
+ * Utility class that provides {@link VarHandle} instances for efficient
+ * multi-byte primitive writes to byte arrays. Falls back gracefully
+ * if VarHandles are not available (e.g., on some Android runtimes).
+ */
```
cowtowncoder, line 19: `@since 3.3`

cowtowncoder:
@pjfanning Any chance for brief jmh perf verification?

pjfanning:
@cowtowncoder I added a benchmark to https://github.com/pjfanning/double-reader-writer

Got these results on my laptop (Java 25)
```
Benchmark                                             Mode  Cnt     Score     Error   Units
CBORGeneratorBenchmark.cborBaselineWriteDoubleArray  thrpt    3   870.669 ± 247.838  ops/ms
CBORGeneratorBenchmark.pr752CborWriteDoubleArray     thrpt    3  1326.809 ±  15.509  ops/ms
```
cborBaselineWriteDoubleArray is current 3.x code
pr752CborWriteDoubleArray is with this PR change

PR #756, opened 2026-08-21
Safer varhandle check (part of #752 fix)

pjfanning:
On Android, VarHandle class is not available. The CBORVarHandleUtil class is likely not load at all because of the import.
This PR adds a try/catch in the calling code so it can live with CBORVarHandleUtil not being loaded.

Relates to #752 

fyi @cowtowncoder


cowtowncoder (approved):

cowtowncoder:
@pjfanning I wish it was easier to test to know for sure... but just to be safe this seems reasonable. Will do one pass with CC first but will likely merge.

**EDIT**: claude code review confirms likelihood of hard failure & soundness of fix. Will merge.

PR #757, opened 2026-08-21
Use SWAR in `SmileParser`

pjfanning:
Similar change to #756 and #752

I added a benchmark to https://github.com/pjfanning/jackson-bench
```
Benchmark                                                  Mode  Cnt     Score    Error  Units
SmileDoubleArrayBench.benchParseFromByteArray             thrpt    5   455.498 ± 32.719  ops/s
SmileDoubleArrayBench.benchParseFromByteArraySwar         thrpt    5   473.796 ± 21.485  ops/s
SmileNameDecodeBench.benchParseFromByteArray              thrpt    5  1227.329 ± 29.590  ops/s
SmileNameDecodeBench.benchParseFromByteArraySwar          thrpt    5  1327.239 ± 28.582  ops/s
```

### DIFF
--- a/smile/src/main/java/tools/jackson/dataformat/smile/SmileParserBase.java
+++ b/smile/src/main/java/tools/jackson/dataformat/smile/SmileParserBase.java
@@ -768,6 +768,39 @@ public abstract class SmileParserBase extends ParserMinimalBase
     }
 
     /**
+     * Helper method for decoding the trailing 1 - 4 bytes of an Object property
+     * name into a right-aligned, zero-padded "quad": that is, produces the same
+     * value as accumulating {@code len} bytes with 8-bit shifts would.
+     *<p>
+     * Reads all 4 bytes with a single load where it can, which may read up to 3
+     * bytes past {@code offset+len}. Those bytes are shifted out and cannot
+     * affect the result, but the read still has to stay within the array: hence
+     * the length check, which also selects the byte-shifting path for a name
+     * that ends within 3 bytes of the end of the buffer.
+     *<p>
+     * Caller MUST have verified that {@code len} bytes are readable at given
+     * offset, and that {@code len} is between 1 and 4.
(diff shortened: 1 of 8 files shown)


smile/src/main/java/tools/jackson/dataformat/smile/SmileParserBase.java
```diff
@@ -30,6 +30,27 @@ public abstract class SmileParserBase extends ParserMinimalBase
...
+    private final static boolean _VARHANDLE_AVAILABLE = _checkVarHandleAvailable();
+
+    private static boolean _checkVarHandleAvailable() {
+        // NOTE: this call is what first loads `SmileVarHandleUtil`, and that class
+        // names `VarHandle` in its field/method signatures. On a runtime without
+        // `java.lang.invoke.VarHandle` (some Android builds) loading it raises
+        // `NoClassDefFoundError` -- an Error, not an Exception -- so `Throwable`
+        // is what has to be caught here.
+        try {
+            return SmileVarHandleUtil.isAvailable();
```
cowtowncoder, line 48: I think it'd make sense to put Error handling within `SmileVarHandleUtil` to minimize code on this side.

pjfanning: I opened https://github.com/FasterXML/jackson-dataformats-binary/pull/756 because it is unsafe to assume that we can load a class with varhandle class imported.

pjfanning: Animal Sniffer in jackson-core is set up to enforce Android 34 min. Googling seems to indicate that should support VarHandles but AnimalSniffer rejected my VarHandle code in https://github.com/FasterXML/jackson-core/pull/1656/ and I had to introduce these workarounds.

PR #772, opened 2026-09-07
Fix `writeString(char[],int,int)` writing enum values twice

pjfanning:
The `EnumLookup` sizing fix that was originally in this PR has been split out into #773.

### `ProtobufGenerator.writeString(char[],int,int)` writes enum values twice

```java
if (_currField.wireType != WireType.LENGTH_PREFIXED) {
    _writeEnum(new String(text, offset, clen));
    // no return -- falls through
}
```

The `String` overload returns after `_writeEnum()`; the `char[]` one did not, so it emitted the enum index and then fell through and emitted the same value **again** as a length-prefixed String. Writing `"WEB"` to an enum field produced 7 bytes instead of 2, i.e. corrupt output.

Pre-dates the 2.x -> 3.x rename and is present on `2.21` too; targeting `3.x` for now, happy to redo against `2.21` and merge forward if you prefer.

### Test

`WriteEnumAsCharArrayTest` — the `char[]` and `String` overloads must produce identical bytes for an enum-valued field, and the result must round-trip. Confirmed it fails with `expected: <2> but was: <7>` if the `return` is removed again.

Full `protobuf` module test suite passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/protobuf/src/main/java/tools/jackson/dataformat/protobuf/ProtobufGenerator.java
+++ b/protobuf/src/main/java/tools/jackson/dataformat/protobuf/ProtobufGenerator.java
@@ -896,6 +896,7 @@ public class ProtobufGenerator extends GeneratorBase
         }
         if (_currField.wireType != WireType.LENGTH_PREFIXED) {
             _writeEnum(new String(text, offset, clen));
+            return this;
         }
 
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Superceded by #776 (backport), closing.

PR #791, opened 2026-09-15
Fix `IonFactory` resource cleanup on failed construction

### LINKED ISSUE #780: `IonFactory` leaks stream when parser/generator construction fails for `File`/`Path`
`IonFactory` extends `DecorableTSFactory` directly (it can use either textual or binary format), so it does not inherit the `File`/`Path` create methods from `TextualTSFactory`/`BinaryTSFactory` — and therefore not the leak fix from FasterXML/jackson-core#1693 either. Its own copies open the stream and hand it off with no `try`/`catch`:

```java
// IonFactory.java:252
public JsonParser createParser(ObjectReadContext readCtxt, File f) {
    final InputStream in = _fileInputStream(f);
    IOContext ioCtxt = _createContext(_createContentReference(f), true);
    return _createParser(readCtxt, ioCtxt, _decorate(ioCtxt, in));
}
```

Anything that throws after the open — a user `InputDecorator`, or `_createParser` itself — leaks the file descriptor. Also note `_fileInputStream(f)` runs *before* `_createContext(...)`, so a failure in `_createContext` leaks too; jackson-core opens the stream after.

Affects `createParser(File)` (:252), `createParser(Path)` (:260), and on the generator side `createGenerator(File, JsonEncoding)` (:358) and `createGenerator(Path, JsonEncoding)` (:366), which are the equivalent of FasterXML/jackson-core#1692.

Fix is the same shape as jackson-core: create the `IOContext` first, then open inside a `try`, and route `RuntimeException` through `DecorableTSFactory._closeOnFailedConstruction(Closeable, RuntimeException)` — `protected` as of jackson-core 3.1, so `IonFactory` can call it directly.

Every other backend (Smile, CBOR, Avro, Protobuf, XML, YAML, CSV) inherits the base-class methods and is already covered; Ion is the only one that opts out.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XrYRabZTvG9TjdXTCghvaM

Dongnyoung:
Fixes #780

Fixes `IonFactory` resource cleanup when parser or generator construction fails after opening `File`/`Path` resources.

`IonFactory` handles these paths separately from the base factory implementations, so the failed-construction cleanup in the base factories does not cover these Ion-specific paths.

### Parser cleanup

There are two failure points that require different cleanup handling:

1. **Failure before an `IonReader` is created**, for example during input decoration.
   - The `InputStream` opened by `IonFactory` is still owned by the outer `File`/`Path` construction path.
   - If construction fails here, that stream needs to be closed directly.

2. **Failure after an `IonReader` is created**, while constructing the `IonParser`.
   - At this point, resource management has been transferred to the `IonReader`/parser construction path.
   - If parser construction fails, the `IonReader` is closed there, which also closes the factory-created input stream.

The initial implementation used a simple `try/catch` around the outer `File`/`Path` path and closed the opened stream on any failure. However, this could close the same stream again when the failure occurs after the `IonReader` has already taken ownership and been closed.

The implementation therefore tracks whether ownership has been transferred to the downstream parser construction path. The outer failure handler only closes the stream if that transfer has not happened yet.

`IOContext` instances created during parser construction are also released on failure so that their `BufferRecycler` leases are returned.

### Generator cleanup

The `File`/`Path` generator paths have the corresponding cleanup issue: if generator construction fails after `IonFactory` opens the output stream, the factory-created stream and its `IOContext` need to be cleaned up.

### Changes

- Close factory-created input streams when input decoration fails before ownership is transferred.
- Close the `IonReader` when parser construction fails after it has been created.
- Avoid closing the same input resource again from the outer `File`/`Path` failure handler after ownership has been transferred.
- Close factory-created output streams when generator construction fails.
- Release `IOContext` instances on failed parser/generator construction.
- Add regression tests for `File` and `Path` parser/generator failure paths.

### Verification

```text
.\mvnw.cmd -pl ion -DskipITs -DskipCheckstyle -Dtest=IonFactoryFailedConstructionTest "-Dsurefire.useModulePath=false" test

### CODE REVIEW

File: ion/src/main/java/tools/jackson/dataformat/ion/IonFactory.java [hunk 1]
```diff
@@ -250,20 +250,42 @@ public Class<IonWriteFeature> getFormatWriteFeatureType() {
...
+        try {
+            in = _fileInputStream(f);
+            in = _decorate(ioCtxt, in);
+            inputCleanupDelegated = true;
+            return _createParser(readCtxt, ioCtxt, in, true);
+        } catch (RuntimeException e) {
+            if (!inputCleanupDelegated) {
+                _closeOnFailedConstruction(in, e);
+            }
+            _releaseContextOnFailedConstruction(ioCtxt, e);
```
cowtowncoder: Would this not work in `finally` block?

cowtowncoder:
Good! Just needs release notes.

ion/src/main/java/tools/jackson/dataformat/ion/IonFactory.java
Dongnyoung: Good point, thanks! 😊😊

